PR #2, opened 2016-06-07
CMake Project File + Linux Build Fixes

Ragora:
This is actually currently unfinished right now, but part of the purpose of this pull is to show the current problems I am having with this.

- There is asserts in the OpenCL renderer that use an unknown method 'get', these are currently just commented in my branch.
- There seems to be an API clash somewhere on FMod. The project readme says FMod Ex, but I am currently getting stuff like ‘FMOD_CHANNEL_FREE’ was not declared in this scope. What version of FMod should be used? Currently the offending code is commented in my branch. I also had to create a softlink in my FMod include dir to suit the FMOD/fmod.h include requirement (I didn't want to just change this to suit my own system) as the version of FMod I have doesn't do this.
- There is some minor adjustments to still be made to the CMakeLists.txt regarding finding SDL2 versus SDL, but that shouldn't be terribly difficult.
- It depends on FindOpenCL which wasn't present in my system version of CMake (2.8.12.2), where it is present in 3.6. I'm not sure where in between it was implemented, but for now the CMake process here forces 3.6.

Right now, it does build on my Linux Mint AMD64 machine using GCC 4.8.4, but I have no real way of testing it right this second. I'll get to brushing up this PR when I can.


afritz1:
Hi Ragora. 

Thanks for taking a look at my code and trying to get it to work on Linux. I'm a bit overwhelmed by the attention my project is getting all of a sudden today, so I'm trying to organize my thoughts at the moment. I am going to be rewriting some of the Readme in a while, too.

Personally, I would've rather done this project in C#, but the native interface with SDL2, FMOD Ex, and OpenCL was too nice to pass up.

I don't have experience with using CMake, but from what I understand it's for using the same compilation rules on different operating systems. I'm new to open-source projects, sorry!

The SDL2, FMOD, and OpenCL libraries I am using are 64-bit.

I am using OpenCL 2.0 in the code, but I believe my graphics card (AMD Radeon HD 7970) is only compatible with OpenCL 1.2. It doesn't cause any problems, though.

My SDL2 version is 2.0.4, and SDL2_image is 2.0.1. 

My SDL2 lib folder has:
- libpng16-16.dll
- SDL2.dll
- SDL2.lib
- SDL2_image.dll
- SDL2_image.lib
- SDL2main.lib
- zlib1.dll

My FMOD lib folder has:
- fmodex64.dll 
- fmodex64_vc.lib (The "_vc" is for Visual Studio I think)

My CL lib folder has:
- OpenCL.dll
- OpenCL.lib

In the executable folder, all of those .dlls are in there, too (which is the usual, right?).

I'll try and figure out what those errors are that you're getting. I'll put the current data folder onto something like Dropbox in the near future so you can actually run the program.

afritz1:
In Rendering/CLProgram.h, the include should be `<CL/cl2.hpp>` instead of `<CL/cl.hpp>`. That's probably why the OpenCL "get" methods are not being recognized. Both cl.hpp and cl2.hpp are in my include directory, but only the second one is being used since the project targets OpenCL 2.0. 

On a side note, it's mentioned in the fourth paragraph of the introduction in cl2.hpp that it is not backward compatible with cl.hpp.

Ragora:
Yea, I did have to change the include for it to build as I don't have a cl2.hpp. It does build without the asserts though, so your current code seems like it should work on either version without the asserts. The whole program does build and link completely in my branch, I just don't have any game data to launch it against. I grabbed the latest OpenCL SDK from AMD's website and I'm using a Radeon R7 360 here.

afritz1:
My FMOD Ex files are from the FMOD Ex Programmer's API found [here](http://www.fmod.org/download-previous-products/#FMODExAPIDownloads), not the FMOD Studio Programmer's API you probably got from [here](http://www.fmod.org/download/#StudioAPI).

I did a search in my fmod.h file for FMOD_CHANNEL_FREE and it's right in there. The version number near the top says `#define FMOD_VERSION 0x00044460`, which translates to 4.44.60, so that means my FMOD version is from Jan 2016.

Ragora:
Will try that right now.

afritz1:
I'm surprised you used std::make_pair instead of the std::pair constructor when I use std::pair all over the place. GCC doesn't like std::pair<std::string, FMOD_SOUND*>? It's fine if it's actually just your preferred style.

[Here](https://www.dropbox.com/s/xc8llh52eahaofs/OpenTESArena_data.zip?dl=0) is the Dropbox data. I only include data in the folder that the program currently looks for, so there's a lot more that's just sitting in my various folders to eventually use.

Let me know what you think of the program if you can run it. There are several placeholders like in the attribute screen for example. There are no character creation questions yet either. The game world is just an infinite space right now, but pressing Tab will bring up the character sheet. The pause screen is a placeholder too, since I considered maybe redesigning it.

Ragora:
make_pair is just what I've been using. I could use that instead if desired. I'll try running it now and see what it does.

afritz1:
Let's merge your branch so I can see your changes working on Windows.

PR #6, opened 2016-06-08
Fix scrollIndex not being initialized

### LINKED ISSUE #5: Sometimes the Player Classes Load
I've been fighting this a lot. The player classes dialog will sometimes show up with no classes to select, this includes if you navigate back (hitting escape) and going back to new game when it already worked.

Ragora:
Fixes #5 

The scrollIndex was sometimes getting ridiculous values because it was getting whatever value happened to be sitting in memory there. Looks like there's probably various other initialization issues, as well.

### DIFF
--- a/OpenTESArena/src/Interface/ListBox.cpp
+++ b/OpenTESArena/src/Interface/ListBox.cpp
@@ -14,7 +14,7 @@
 
 ListBox::ListBox(int x, int y, FontName fontName, const Color &textColor, int maxDisplayed,
 	const std::vector<std::string> &elements, TextureManager &textureManager)
-	: Surface(x, y, 1, 1), textureManagerRef(textureManager)
+	: Surface(x, y, 1, 1), textureManagerRef(textureManager), scrollIndex(0)
 {
 	assert(maxDisplayed > 0);
 


afritz1:
Ah! Yes, I forgot to initialize the `scrollIndex` variable. In debug mode in Visual Studio, I assume it always defaults to 0 because the first class (Acrobat) is always at the top of the list every time, so I never noticed that it was uninitialized. I should've put some asserts in there.

With the pull request I just merged before this, I'm now getting an error due to some Windows namespace being implicitly included into GameWorldPanel.cpp, so that interferes with the Rectangle class at line 219 in `compassSlider.blit()`.

Also, if I include just `<CL/cl.hpp>` and not `<CL/cl2.hpp>` in CLProgram.h, I get errors about `clCreateSampler` being deprecated.

Ragora:
I added a couple of standard headers around to include the necessary functionality for it to compile here, interesting that it's causing problems due to inclusion.

afritz1:
When I change `<CL/cl.hpp>` to `<CL/cl2.hpp>` in CLProgram.h, the Windows namespace clash with Rectangle goes away and the program builds correctly. Strange how cl.hpp affects that.

I haven't been using the CMake file yet.

I think it would be for the best if that header in CLProgram.h remains as the OpenCL 2.0 version `<CL/cl2.hpp>`.

Ragora:
Would just need to obtain an older SDK on my end, should be fine.

PR #20, opened 2016-06-14
Implement music volume handling, and a couple MIDI fixes

kcat:


afritz1:
In the `OpenALStream()` initialization list, I moved the `mBuffers` member outside of it because Visual Studio's compiler does not have array initialization implemented in the initialization list yet.

kcat:
Okay, added a workaround for that.

PR #21, opened 2016-06-16
Read the default palette

kcat:
This will help with reading the original 8-bit image files.


afritz1:
It's probably fine, but you can use my Color class instead of PalColor if you'd like. They appear to be structurally equivalent, only the Color class isn't just a typedef of a std::array. It's in the Media folder if you're interested.

kcat:
Didn't notice that. Done.

PR #30, opened 2016-06-18
Implement type 8 IMG decompression

kcat:


afritz1:
Tell me about the decode08Type function. It looks pretty busy!

kcat:
The comments describe it as best I can, honestly. I found code for handling the algorithm here:
https://github.com/Dysperia/BSATool/blob/master/src/various.cpp
then as I rewrote it, I tried to understand exactly what it was doing. As best I can tell, it's some form of adaptive [Huffman coding](https://en.wikipedia.org/wiki/Huffman_coding) with LZ compression, creating notable similarities to [DEFLATE](https://en.wikipedia.org/wiki/DEFLATE). It may actually be DEFLATE, just without the 3-bit header.

Essentially it starts with a tree of nodes (```NodeTree```), with index 626 being the root, the value of which being the base index for two branches (value = branch1, value+1 = branch2). Each branch itself has two branches recursively down to leaf nodes. There's 314 leaf nodes, for 314 encoded values. Each bit of the input stream indicates how to traverse down the node tree, until it reaches a leaf node (indicated by a node value of ```>= 627```, and ```original_value = leaf_node_value - 627```).

Once a leaf node is reached, it increments the use-count (aka frequency, ```NodeFreq```) for the nodes along the branch and resorts the tree. The idea being that the more a leaf is used, the closer to the root of the tree it becomes so it takes less bits to keep referring to the same leaf. These two stages are what make me think of adaptive Huffman coding, since it's the same principle.

The value retrieved from the leaf node is either a direct output pixel (if ```<256```), or is used with an extra 8 bits from the input to duplicate a previous run of pixels. This is similar to LZ compression.

It keeps reading bits from the input, traversing the tree to find leaf nodes, and generate output pixels using the leaf node values, until the output is fully generated.

PR #31, opened 2016-06-18
Handle more IMG files

kcat:
Also replaces some texture file references to Arena's. I'm not sure where ```interface/parchment/parchment_popup``` comes from, though. And ```TAMRIEL.IMG``` is either not 100% correct for the race select screen, or it's supposed to use palette trickery to hide the yellow dots (we'd need to load a separate version that overrides its built-in palette for conversion), but it's good enough.


afritz1:
I screen-captured `interface/parchment/parchment_popup` in DOSBox and erased the text. I think the original is actually a repeating pattern, not just a single image.

Issue #17, opened 2016-06-12
Building VFS static lib


afritz1:
How should I approach turning the components files into a static .lib in a portable manner? Should it be part of the CMake file so a Windows developer can build it automatically?

**Edit:** It looks like the static library instructions are already in CMakeLists.txt, but my CMake doesn't know where to find OpenAL at. Is there a standard "FindOpenAL.cmake" that should be added to the repository?

Ragora:
To build components into static libs, you'd have to make sub CMake projects for them and use ADD_SUBDIRECTORY. That's at least how I did it here: 

https://github.com/DraconicEnt/KGE/blob/develop/CMakeLists.txt#L122

So for the "support" component:

https://github.com/DraconicEnt/KGE/blob/develop/components/CMakeLists.txt#L3
https://github.com/DraconicEnt/KGE/blob/develop/components/support/CMakeLists.txt

afritz1:
I guess I'm just encountering some issues because I'm not sure what the best CMake conventions are. The goal is to have the CMakeLists.txt file produce the desired output for both your Linux IDE and my Visual Studio.

In general, what is a good directory to have the include and lib folders for the project in? For example, if I was a newcomer and I cloned the repository to a local folder, should the include and lib folders also be in that same place used with a .gitignore, or should they be in a nearby folder that other projects might refer to? I don't want to put my libraries in the `System32` directory for example because those are harder to keep track of.

Is there a way to make the generated project file keep my folder structure as well? Visual Studio just seems to be throwing all the code files into a big list.

kcat:
> To build components into static libs, you'd have to make sub CMake projects for them and use ADD_SUBDIRECTORY.

I haven't had to do that. I've never had a problem using ADD_LIBRARY to build a static library, ADD_EXECUTABLE to build an exe, then specify that library to link with the exe, all in one cmake project.

For OpenAL, the easiest way to get building with it is to download [Creative's core SDK](http://openal.org/downloads/OpenAL11CoreSDK.zip) and run the installer. It sets specific registry keys during installation about where it's installed to, and the FindOpenAL.cmake script is actually smart enough to check them to find the appropriate lib and header locations. And even though it's built against a different set of libs, OpenAL Soft is 100% ABI compatible to use with the resulting exe.

Ragora:
That's just for logical separation.

afritz1:
[...]

PR #35, opened 2016-06-22
CMake: Use the CMake way to select C++11 standard

pcercuei:
Signed-off-by: Paul Cercueil <paul@crapouillou.net>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,8 +10,6 @@ set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/")
 
 # Configure for GNU Compiler
 IF (CMAKE_CXX_COMPILER_ID STREQUAL "GNU" OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang")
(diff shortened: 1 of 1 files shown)


afritz1:
Thanks for the change, @pcercuei. I'm still a little unsure about how to correctly use CMake yet. I don't quite have it working the way I'd like with Visual Studio. It's talked about in [this issue](https://github.com/afritz1/OpenTESArena/issues/17#issuecomment-226917415).

kcat:
Unfortunately this breaks builds on older cmake versions. Debian is currently on cmake 3.0.2, and now building fails because it's not compiling with ```-std=c++11``` or equivalent anymore.

Ragora:
And I've got 2.8.12 at the latest on Mint.

pcercuei:
Seriously? CMake is at 3.5.2 now, you guys should update your systems :D
Maybe revert it for now then.

Ragora:
@pcercuei My system is running the latest stable.

> ragora@Kiaro:~ > lsb_release -a
No LSB modules are available.
Distributor ID: LinuxMint
Description:    Linux Mint 17.3 Rosa
Release:        17.3
Codename:       rosa

https://www.linuxmint.com/download.php

There is a beta 18, but I try to just use stables.

PR #36, opened 2016-06-22
CMake: Split the main CMakeLists.txt into separate files

pcercuei:
Plus some cleanups.

Signed-off-by: Paul Cercueil <paul@crapouillou.net>


afritz1:
In ` OpenTESArena/CMakeLists.txt`, shouldn't the word `TESArena` be `OpenTESArena`? I'm not sure why it was shortened in the first place.

Ragora:
I shortened it to prevent a naming clash with the OpenTESArena folder name.

afritz1:
Okay. In release packages, though, I'd like the executable to use the full name.

Ragora:
You'd have to change the folder name or change casing somewhere (which I personally wouldn't particularly like).

afritz1:
I mean for the `.tar.gz` release package that would just have the executable and other assorted files in it, so people are running `OpenTESArena` instead of `TESArena`. This wouldn't be a part of the build itself; just what people get at the end.

Ragora:
The thing I linked would be a tool to generate those automatically (which also includes App Bundles, RPM's and Debs) from the CMake projects. That's why I brought it up.

afritz1:
Oh, okay. We can use package creation tools at some point then.

Ragora:
I haven't used CPack myself yet, I'm sure someone would have a better idea of its any good.

components/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+PROJECT(components CXX)
+
+FILE(GLOB COMP_SOURCES vfs/*.h vfs/*.hpp vfs/*.cpp
+	archives/*.h archives/*.hpp archives/*.cpp
+)
+
+IF (WIN32)
+	FILE(GLOB COMP_MISC_SOURCES misc/*.h misc/*.hpp misc/*.cpp)
```
pcercuei: OOps, there's a bug here. Sorry about that.
It should be COMP_WIN32_SOURCES, which is used line 11.

PR #37, opened 2016-06-23
Fix build for older cmake, use more images from Arena's data

kcat:


afritz1:
A problem with using `TAMRIEL.MNU` now for race selection is that it has the "Exit" text at the bottom right, which is only shown in the in-game world map. I should be able to just blit a chunk of `TAMRIEL.IMG` (using TextureName::WorldMap) over the bottom right corner to cover it up.

kcat:
> A problem with using TAMRIEL.MNU now for race selection is that it has the "Exit" text at the bottom right, which is only shown in the in-game world map.

Interesting, I didn't notice that. That makes me curious about what this is supposed to be doing. The race select screen in vanilla does lack the "Exit" text the MNU file has, though it also lacks the yellow dots the IMG file has. Unfortunately I can't run my way out of Arena's starting dungeon to get a look at its world map to see what it looks like.

afritz1:
The in-game world map has "Exit" at the bottom right and no yellow dots. I'm figuring out how to cover up the "Exit" in the race selection right now.

kcat:
> The in-game world map has "Exit" at the bottom right and no yellow dots.

Perhaps it'd be better to swap the uses of TAMRIEL.MNU and TAMRIEL.IMG, then. That is, use the MNU for the world map menu and the IMG for the race select. We can figure out the yellow dots later.

PR #41, opened 2016-07-28
Fix building on GCC/Clang and fix song looping

kcat:


afritz1:
I wonder why the `PaletteName` would be considered an incomplete type, since in C++11, enum classes can be used even when forward declared because their underlying type is assumed to be `int`.

kcat:
```Surface``` was the incomplete type, so it was choking on the ```std::unordered_map``` declaration for it since it didn't know how to allocate one. It might've been due to something else, since I've generally not had an issue declaring STL containers with incomplete types, ones that essentially just store a pointer to the object type instead of an instance of the object type, at least, as long as no methods are called that requires knowing its size or other relevant properties (i.e. making sure there's no inlined or auto-generated constructors, destructors, etc that will imply code that needs the fully defined type), but I didn't look too deep into it.

The enum issue is that ```std::hash``` couldn't take one as a type before C++14, and though I could've just manually casted it to ```int``` or something, I didn't see the point since a hash's purpose is to give a unique integer value for a unique object value. But if the object type is already an integer, a unique object value is implicitly a unique integer value.

afritz1:
Yeah, that makes sense. 

I might trim out some of the `Surface` things later on since the rendering is now much more focused on using `SDL_Texture` for hardware-accelerated rendering. There are a few things like text boxes that still use surfaces, but I'd like to change them to `SDL_Texture` eventually. I think the `std::unordered_map` for surfaces might still hang around in the texture manager, though, for easy conversion of surfaces to OpenCL float4 buffers.

PR #43, opened 2016-08-24
Linux compile fix

mdmallardi:
This pull  request allows compilation on Linux with g++


afritz1:
Thanks. I don't have a Linux computer yet, and Visual Studio's compiler is a bit more permissive, so some of these things get missed. I've been thinking about starting Travis CI sometime to cover that base. 

And about that `vector` include, it looks like MSVC includes `vector` through `<unordered_map>` for some reason.

PR #45, opened 2016-08-29
Fix problems in the OpenCL find script

Ragora:
1. Removes OpenCL 2.0 from the search test
2. Simplifies the library search (we don't really need to search based on machine architecture)
3. Add a few extra possible search directories: /usr/lib/x86_64-linux-gnu and /usr/lib/i386-linux-gnu
4. Add .so.1 as a possible file suffix for libOpenCL because packages might not install libraries that have a softlink or library without the suffix, like: http://packages.ubuntu.com/trusty/amd64/nvidia-libopencl1-331/filelist
5. Search specifically for cl2.hpp and fail if it can't be found, also allow the directory for it to be specified if it's not found in the OpenCL install (so we can store it elsewhere)
6. Fixes the way include files are searched for (we really only want to know the root of where the  includes are, but searches within the bigger SDK's may be broken without the prefix, but I don't know what they typically look like right now either)

This should make the CMake script work more reliably, as it failed completely to find the library file if you're using something like the package above to provide libOpenCL (as it never searches /usr/libx86_64-linux-gnu or /usr/lib/i386-linux-gnu). Also it specifically checks for cl2.hpp to prevent further hear ripping when builds fail because you have regular OpenCL but not the C++ bindings from here (the file works, just needs renamed): https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/input_cl2.hpp

Not sure if any package provides cl2.hpp, but it definitely didn't come by default in the package I linked above and it can easily be yanked from that Github URL.

Also, the version checking is totally pointless right now as none of the scripts really make use of it. We can export version data to C++ visible defines for conditional compilation of OpenCL code, but right now it's totally unused.

### DIFF
--- a/cmake/FindOpenCL.cmake
+++ b/cmake/FindOpenCL.cmake
@@ -38,7 +38,7 @@ function(_FIND_OPENCL_VERSION)
   set(CMAKE_REQUIRED_QUIET ${OpenCL_FIND_QUIETLY})
 
   CMAKE_PUSH_CHECK_STATE()
-  foreach(VERSION "2_0" "1_2" "1_1" "1_0")
+  foreach(VERSION "1_2" "1_1" "1_0")
     set(CMAKE_REQUIRED_INCLUDES "${OpenCL_INCLUDE_DIR}")
 
     if(APPLE)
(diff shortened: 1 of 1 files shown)


afritz1:
I tried using `<CL/cl.hpp>` but I was getting errors about `clCreateSampler`, so I haven't found a way to avoid needing `cl2.hpp` yet.

afritz1:
I'm not sure if I'm reading the changes right, but the search paths should still look in the `CL` folder for `cl2.hpp`. On my computer, it's stored in `C:\Program Files (x86)\AMD APP SDK\3.0\include\CL`, since there's also a `GL` folder right next to `CL` in the `include` folder.

Ragora:
There's a bunch of default paths that were there since before I even messed with it, we can add more if necessary and the environment variables should set be set when your appropriate SDK is installed. I just might've broken the searching for them as noted in number 6 up there but I haven't been able to check what suffixes might be required for those SDK's yet.

PR #46, opened 2016-08-29
Set up Travis-CI build

Allofich:
This gets Travis-CI through compiling OpenTESArena on gcc.

Currently WildMIDI isn't being found.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+language: cpp
+sudo: required
+dist: trusty
+os: linux
+compiler: gcc
+
+before_install:    
+- sudo apt-get install libsdl2-dev
+- sudo apt-get install libsdl2-image-dev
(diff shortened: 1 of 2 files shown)


afritz1:
Thanks for figuring that out. I'm on Windows, and since there were some Linux issues that got through recently because MSVC is more permissive, I thought it'd be good to have Travis CI for better coverage on that part.

Allofich:
Sure. I'm on Windows, too, and I'm not familiar with Unix commands, so that was a bit of a challenge here, too.

Looking at the Travis config it looks like it automatically tacks on "Group: Stable" at the end. If you want it to look nicer / be explicit about the configuration, you could add that to the .travis.yml at the top I guess, if you care about such things.

afritz1:
I see it. Yeah, why not. I'll add that shortly.

PR #47, opened 2016-08-29
Correct the find OpenCL for both cases

Ragora:
This should make the FindOpenCL script work in my situation and the one in the Travis CI. Basically just shotgunned it because I still have no real idea what the SDK folder structure looks like, but from the discussion in #46 it seems to line up with what I added here.

### DIFF
--- a/cmake/FindOpenCL.cmake
+++ b/cmake/FindOpenCL.cmake
@@ -75,7 +75,7 @@ endfunction()
 
 find_path(OpenCL_INCLUDE_DIR
   NAMES
-    cl.h cl.h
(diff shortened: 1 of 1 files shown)


Ragora:
Looks like it's fine then, these changes just make less assumptions of what the directory structure actually looks like and focuses on performing checks against the possibilities. (the suffixes are appended to all PATHS specified to check potential matches, as well).

Allofich:
Hm, I know you said the versions printed by the find script are irrelevant, but it looks like after this PR the log is again saying "not found" for the CL versions. No problem there?

PR #52, opened 2016-09-30
Add CMake artifacts, and linux binary to .gitignore.

mdmallardi:
This commit adds all current CMake artifacts and the final Linux binary to .gitignore, so these will no longer be attempted to be indexed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@
 *.exe
 *.out
 *.app
+TESArena
 
 # Unnecessary Visual C++ files
 ipch/
@@ -59,3 +60,9 @@ options/
 *.tar.gz
 *.zip
 *.7z
+
+# Cmake artifacts
+CMakeCache.txt
+CMakeFiles/
+Makefile
+cmake_install.cmake


afritz1:
I've been told by @Ragora before that the Linux executable has to be named something other than `OpenTESArena` to avoid a name collision with something else (presumably the `OpenTESArena` folder)? However, I would like to use the full name for the executable at some point just for consistency with the project name. This is not an issue on Windows, though.

I assume that GCC drops the executable in the top directory then? Maybe it should go in a build directory instead to fix this name collision? What are your thoughts?

mdmallardi:
The cmake build system does drop the binary right in the root directory, calling it 'TESArena', instead of OpenTESArena. (Hence why I have that ignored)  I did forget one CMake artifact that should be ignored: 'install_manifest.txt' (Only generated when someone runs 'make install')  I personally think that it's fine, except for the fact that I think everyone would prefer an amount of consistancy between platform builds with the binary naming conventions.  In that case, certainly dropping the binary into a build/ subdirectory, and then just ignoring that subdirectory would be preferable.

afritz1:
Maybe the addition of a build directory could be done in another pull request, and the .gitignore can be revised again at that time. I'd do it myself if I knew more about accepted CMake practices, but I don't at the moment.

PR #53, opened 2016-10-03
Removed dependency on SDL2_image library from CMake build system.

mdmallardi:
This removes the SDL2_image dependency from the Unix builds as well.


afritz1:
Thanks!

PR #55, opened 2016-10-25
Update .travis.yml

Ragora:
Package repos will be lagging behind on SDL 2.0.5 for a while (as we now require some functionality specific to 2.0.5 in TESArena now), so we clone via mercurial the 2.0.5 release and build it.  It should work, but we won't fully know until we merge and have Travis test within its environment. 

This is why the last Travis build failed -- we pulled SDL2 from repos which pulled down some version < 2.0.5 which will not include the functionality that was recently depended upon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,14 @@ os: linux
 compiler: gcc
 group: stable
 
-before_install:    
+before_install:
+  - sudo apt-get install mercurial
+  - hg clone http://hg.libsdl.org/SDL -r release-2.0.5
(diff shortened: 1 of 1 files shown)


Ragora:
Also, we can use -j# on the make command for SDL to speed up that process depending on how Travis wants us to use that functionality.

afritz1:
Okay. This will be a temporary fix until SDL 2.0.5 is available in the Ubuntu distribution. I'm not worried about speed of compilation right now. Just correctness.

PR #61, opened 2017-01-15
Create pklite_specification.md

Dozayon:
Added a specification for the PKLITE V1.12 compression format.

### DIFF
--- a/docs/pklite_specification.md
+++ b/docs/pklite_specification.md
@@ -1,0 +1,240 @@
+# PKLITE V1.12 compression format specification
+
+
+This specification should work with any executable compressed with PKLITE V1.12, but it is mainly designed for A.EXE from Bethesda's 'The Elder Scrolls: Arena'.
+
+
+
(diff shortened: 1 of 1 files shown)


afritz1:
Excellent work. I'll see about starting on the implementation either today or in the next couple days.

hackerb9:
@dozayon thank you for this excellent documentation. I'm reusing it in a [standalone decompressor](https://github.com/hackerb9/depklite) based on the keen dreams depklite routines. I have a request which I think may also benefit projects like OpenTESArena. Would you be willing to add to the clean room documentation some error checking? For example:

* What is the method of discovering which version of pklite was used? This documentation seems to work with versions of pklite other than 1.12, by the way, but with possible extra garbage or output corruption.
* How is a corrupt stream detected? Or is that impossible?
* Is there a more general way of discovering the start byte? Or is it hardcoded based on pklite version? If the latter, could you please include a table of offsets?

Thanks!

PR #63, opened 2017-01-21
Change opentesarena.rc encoding to UTF-8

Dozayon:
Change the encoding of opentesarena.rc from UTF-16LE to UTF-8.
This makes it compatible with more compilers.


afritz1:
Okay!

PR #64, opened 2017-02-01
Fix EOL check in text parsers

Dozayon:
The default behavior of `getline` is to get the next line without the
EOL symbol(s). Hence, `line.at(0)` should in theory always throw an
exception for an empty line, since there would be no characters in the
string.

However, this doesn't happen on Windows. The EOL sequence on Windows
is '\r' '\n', but `getline` uses just '\n' as a delimiter, resulting
in '\r' being left in the string, hence making the code work.

The code doesn't work if the EOL sequence of the text files is just
LF. This fix should solve that, while still keeping the code
compatible with CRLF endings on Windows. Checking for '\n' is now
redundant.


afritz1:
Hmm, so `std::getline()` only checks for `\n` then (in the default case). Okay. When I was writing that code, I figured that checking for both characters was the safest, but I understand now.

By the way, the character class parser is just a thing I made up as a placeholder until the classes could be read in from the Arena data.

PR #67, opened 2017-04-28
Address some warnings from CPPCheck

Allofich:
This is just some cleanup, which I enjoy doing, to address warnings that show when checking the code with the CPPCheck code analyzer tool.

One change involved switching the order of arguments in the camera3D constructor from (position, direction) to (direction, position) to match the order the variables are declared. This is only called in one place right now but if you like the current order I could change this back.

Also, I didn't touch these but just in case it is useful for you, there are the following currently unused variables, according to CPPCheck:

Filename           Lines                                     Variable names
IMGFile.cpp      207, 208, 209, 210                 xoff, yoff, width, height
CFAFile.cpp      29, 30, 72                               unknown1, unknown2, dstOffset
DFAFile.cpp      26, 27, 58                               unknown1, unknown2, chunkSize
ItemParser.cpp 36                                          text
Renderer.cpp    360                                       screenHeight
FontFile.cpp     59                                          counts

And the unused private function
NonPlayer.h     34                                          NonPlayer::getAnimationType

### DIFF
--- a/OpenTESArena/src/Entities/Camera3D.cpp
+++ b/OpenTESArena/src/Entities/Camera3D.cpp
@@ -7,10 +7,9 @@
 #include "../Math/Quaternion.h"
 
 Camera3D::Camera3D(const Double3 &position, const Double3 &direction)
-	: position(position), forward(direction)
+	: forward(direction), right(forward.cross(Double3::UnitY).normalized()),
+	up(right.cross(forward).normalized()), position(position)
 {
-	this->right = this->forward.cross(Double3::UnitY).normalized();
-	this->up = this->right.cross(this->forward).normalized();
 }
 
 const Double3 &Camera3D::getDirection() const
--- a/OpenTESArena/src/Entities/Doodad.h
+++ b/OpenTESArena/src/Entities/Doodad.h
@@ -15,7 +15,7 @@ private:
 	Animation animation;
 	Double3 position;
 public:
-	Doodad(const Double3 &position, const Animation &animation,
+	Doodad(const Animation &animation, const Double3 &position,
 		EntityManager &entityManager);
 	virtual ~Doodad();
 
--- a/OpenTESArena/src/Entities/Player.cpp
+++ b/OpenTESArena/src/Entities/Player.cpp
@@ -15,16 +15,10 @@ Player::Player(const std::string &displayName, GenderName gender, int raceID,
 	const CharacterClass &charClass, int portraitID, const Double3 &position, 
 	const Double3 &direction, const Double3 &velocity, double maxWalkSpeed, 
 	double maxRunSpeed)
-	: charClass(charClass), displayName(displayName), camera(position, direction), 
-	velocity(velocity), weaponAnimation(WeaponType::Fists /* Placeholder for now. */) 
+	: displayName(displayName), gender(gender), raceID(raceID), charClass(charClass), portraitID(portraitID), camera(position, direction),
+	velocity(velocity), maxWalkSpeed(maxWalkSpeed), maxRunSpeed(maxRunSpeed), weaponAnimation(WeaponType::Fists /* Placeholder for now. */)
 {
 	assert(portraitID >= 0);
-
-	this->maxWalkSpeed = maxWalkSpeed;
-	this->maxRunSpeed = maxRunSpeed;
-	this->gender = gender;
-	this->raceID = raceID;
-	this->portraitID = portraitID;
 }
 
 Player::~Player()
--- a/OpenTESArena/src/Assets/Compression.h
+++ b/OpenTESArena/src/Assets/Compression.h
@@ -19,7 +19,7 @@ class Compression
 	~Compression() = delete;
 public:
 	// Uncompresses an RLE run of bytes.
-	static void decodeRLE(const uint8_t *src, uint32_t count, 
+	static void decodeRLE(const uint8_t *src, uint32_t stopCount,
 		std::vector<uint8_t> &out);
 
 	// Works with .IMG and .CIF type 4 files.
--- a/OpenTESArena/src/Entities/Doodad.cpp
+++ b/OpenTESArena/src/Entities/Doodad.cpp
@@ -2,9 +2,9 @@
 
 #include "EntityType.h"
 
-Doodad::Doodad(const Double3 &position, const Animation &animation,
+Doodad::Doodad(const Animation &animation, const Double3 &position,
 	EntityManager &entityManager)
-	: Entity(entityManager), position(position), animation(animation) { }
+	: Entity(entityManager), animation(animation), position(position) { }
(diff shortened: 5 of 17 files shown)


afritz1:
I enjoy making sure code looks neat and tidy, too. I'll review the changes.

OpenTESArena/src/Entities/Camera3D.cpp
```diff
@@ -6,8 +6,8 @@
 #include "../Math/Constants.h"
 #include "../Math/Quaternion.h"
 
-Camera3D::Camera3D(const Double3 &position, const Double3 &direction)
-	: position(position), forward(direction)
+Camera3D::Camera3D(const Double3 &direction, const Double3 &position)
```
afritz1: I actually prefer having `position` first in the parameter list. The 3D frame (forward, right, up) is private because those values should only be changed through methods, but the position can be changed directly, so it is public.

OpenTESArena/src/Entities/Doodad.h
```diff
@@ -12,8 +12,8 @@
 class Doodad : public Entity
 {
 private:
-	Animation animation;
 	Double3 position;
+	Animation animation;
```
afritz1: I guess this change is okay. I normally put larger structs first, but that's only really relevant to byte padding which is hardly a concern here.

OpenTESArena/src/Game/Options.h
```diff
@@ -33,7 +33,7 @@ class Options
 	std::string arenaPath; // "ARENA" data path.
 	bool skipIntro;
 public:
-	Options(std::string &&arenaPath, int screenWidth, int screenHeight, bool fullscreen,
+	Options(std::string &&dataPath, int screenWidth, int screenHeight, bool fullscreen,
```
afritz1: The intended name for this parameter is `arenaPath`.

OpenTESArena/src/Interface/CinematicPanel.h
```diff
@@ -22,8 +22,8 @@ class CinematicPanel : public Panel
 	double secondsPerImage, currentSeconds;
 	int imageIndex;
 public:
-	CinematicPanel(Game *game, const std::string &paletteName,
-		const std::string &sequenceName, double secondsPerImage,
+	CinematicPanel(Game *game, const std::string &sequenceName,
+		const std::string &paletteName, double secondsPerImage,
```
afritz1: I didn't notice this bug until now. I wonder why it works. We should match the parameter ordering here with `ImagePanel.h/.cpp` and `ImageSequencePanel.h/.cpp`, where `paletteName` comes before the other strings.

OpenTESArena/src/World/VoxelBuilder.h
```diff
@@ -24,7 +24,7 @@ class VoxelBuilder
 	static std::vector<Rect3D> makeSizedBlock(int x, int y, int z, float y1, float y2);
 
 	// Creates a 1x1x1 block at the given coordinates.
-	static std::vector<Rect3D> makeBlock(int x, int y, int z);
+	static std::vector<Rect3D> makeBlock(int cellX, int cellY, int cellZ);
```
afritz1: This should remain `int x, int y, int z` like all of the other nearby methods.

Allofich:
OK, I'll start making these changes. If you want to do it yourself and just use this PR as a reference, that's fine, too, just say so.

afritz1:
Go ahead and make the changes. I'm looking at the unused variables and determining which ones to keep. I'll change those afterwards.

OpenTESArena/src/Entities/Camera3D.cpp
Allofich: `position` is first again in the parameters, but I initialize `forward,` `right` and `up` first to match the order they are declared to satisfy the warning. Is this all right?

OpenTESArena/src/Entities/Doodad.h
Allofich: I put the declaration back where it was and instead changed the order of initialization to match.

OpenTESArena/src/Game/Options.h
Allofich: Done.

OpenTESArena/src/Interface/CinematicPanel.h
Allofich: I don't know why it works. But I changed it so that `paletteName` is always first.

OpenTESArena/src/World/VoxelBuilder.h
Allofich: Done.

OpenTESArena/src/Entities/Camera3D.cpp
Allofich: If not, it's not a big deal. If I understand correctly the warning is just to keep you from mistakenly thinking the variables are being initialized in the order of the initialization list, when they are in fact initialized in the order of their declaration.

afritz1: Yeah. I might tweak it some in the future depending on how I want to use Camera3D.

OpenTESArena/src/Entities/Player.cpp
```diff
@@ -15,16 +15,10 @@ Player::Player(const std::string &displayName, GenderName gender, int raceID,
 	const CharacterClass &charClass, int portraitID, const Double3 &position, 
 	const Double3 &direction, const Double3 &velocity, double maxWalkSpeed, 
 	double maxRunSpeed)
-	: charClass(charClass), displayName(displayName), camera(position, direction), 
-	velocity(velocity), weaponAnimation(WeaponType::Fists /* Placeholder for now. */) 
+	: displayName(displayName), gender(gender), raceID(raceID), charClass(charClass), portraitID(portraitID), camera(position, direction),
+	velocity(velocity), maxWalkSpeed(maxWalkSpeed), maxRunSpeed(maxRunSpeed), weaponAnimation(WeaponType::Fists /* Placeholder for now. */)
 {
```
afritz1: I usually don't put primitive types (`int`, `float`) in the initializer list because it isn't overriding any default contructor like for classes. I'll change this at some point.

afritz1:
Some of these changes are different from my usual thinking, but maybe that just says that my style could use some more refining. When I bump into those places in the future, I may or may not tweak them a bit. For now, it's okay.

Allofich:
Well, I was looking at the player constructor after your latest comment. It looks like just having the constructor declaration in player.hpp have the parameters in the same order as the variables are declared is enough to stop the warning in CPPCheck, even if the way the initialization is written in player.cpp is returned to how it was.

The whole fourth commit is to address the warnings about initialization lists not matching variable declaration order, so if that's not a concern to you, you could revert it if you want. But at any rate, thanks for merging.

Issue #74, opened 2017-07-07
Add macOS support


Thunderforge:
This project currently has Windows and Linux builds for each release. I'd like to see macOS releases as well so that I can run OpenTESArena on my preferred platform.

afritz1:
You saw what I said on the OpenMW IRC, didn't you? 😉

The best way to get the ball rolling is to have a Mac user try and build the project with CMake, and then leave their findings here. There aren't very many dependencies -- just SDL2, OpenAL Soft, WildMIDI, and a C++11 compiler. WildMIDI is even optional for now if you don't want music (with the `HAVE_WILDMIDI` define, although that might change in the future). Instructions for building from source are [here](https://github.com/afritz1/OpenTESArena/tree/162e95b0e9a14a347716c12422e77ac8c0719404#building-from-source).

There might need to be some changes to the top-level `CMakeLists.txt`, which is fine. We'll probably also want to add Clang to the list of Travis CI builds in `.travis.yml`.

In order to have consistent macOS releases, we'd need someone on that operating system to maintain them, since I'm on Windows and @Ragora is on Linux.

Thunderforge:
Yup, I did see you saying that nobody had asked for a Mac release, so I figured I'd do that.

I'm a complete newbie to C++ (I have plenty of Java and Python experience though) and am especially confused with dependency management (on Java, we just use Gradle or Maven to download and build everything with one command, and in a sandboxed mode too). That said, I decided to give it a try.

First, I installed CMake, OpenAL Soft, and SDL via Homebrew (WildMIDI wasn't available). I simply ran  `cmake -G "Unix Makefiles" ..`, and was able to build successfully (with a warning that I wouldn't have MIDI music). When I ran `make` after that, I got a number of errors like this:

```
~/OpenSource/OpenTESArena/components/archives/archive.cpp:61:19: error: 
      use of overloaded operator '-' is ambiguous (with operand types
      'std::streampos' (aka 'fpos<__mbstate_t>') and 'std::streamsize'
      (aka 'long'))
    return newPos - mStart;
           ~~~~~~ ^ ~~~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/string:483:37: note: 
      candidate function
    _LIBCPP_INLINE_VISIBILITY fpos  operator- (streamoff __off) const {f...
                                    ^
[...]

PR #75, opened 2017-07-09
Bugfix/Add type conversion to long for macOS + Clang compatibility

Thunderforge:
As part of Mac compatibility described in #74.


afritz1:
Looks good to me. Thanks for getting it running on macOS!

PR #77, opened 2017-07-10
Copy the data and options folders via cmake

Thunderforge:
This eliminates the need to "Copy the data and options folders to where the executable is in the build directory" as described in the setup guide, and will be very helpful for macOS app bundling.


afritz1:
This is a good idea. I just didn't know how to do it in CMake right away.

I wonder if this would work with the Visual Studio build structure, since in my `build` folder, there's the primary OpenTESArena solution for the whole project (ALL_BUILD, TESArena, ZERO_CHECK), and then there are folders with each sub-project. I assume it would work, because the default "executable directory" for TESArena is just the project folder itself.

Thunderforge:
I'm a complete CMake newbie, so I can't guarantee that this is the best way, just the way that I found. That said, I think it will work with the structure you describe. Hopefully the branch can be tested by someone with Visual Studio.

afritz1:
Well, I tested the CMake changes in the CMake GUI and it puts the data and options folders one folder above where they should be. I'm _kind of_ anxious about adding more and more platform-specific branches in the `CMakeLists.txt`, so maybe there's a better way, like adding it in the `TESArena` folder specifically?

Thunderforge:
Where should they be? For me using CLion, the structure is this:

 - OpenTESArena
   - OpenTESArena (from GitHub)
      - src
      - CMakeLists.txt
   - data (from GitHub)
   - options (from GitHJub)
   - etc.
   - cmake-build-debug (auto-generated by CLion)
     - data (copied via cmake)
     - options (copied via cmake)
     - TESArena (compiled app)

Do you think that there is [another CMake variable](https://cmake.org/Wiki/CMake_Useful_Variables) that would point to the directory that you are looking for?

afritz1:
Just did some more experiments and it looks like putting the file copying code in `OpenTESArena/CMakeLists.txt` would work better:

```CMake
...
IF (WIN32)
    LIST(APPEND TES_SOURCES ${TES_RESOURCES})
    ADD_DEFINITIONS("-D_SCL_SECURE_NO_WARNINGS=1")
ENDIF()

# Copy over required files.
FILE(COPY ${CMAKE_SOURCE_DIR}/data DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
FILE(COPY ${CMAKE_SOURCE_DIR}/options DESTINATION ${CMAKE_CURRENT_BINARY_DIR})

ADD_EXECUTABLE (TESArena ${TES_SOURCES})
...
```

So no need to change the top-level `CMakeLists.txt`.

Thunderforge:
Good call; I've checked in a commit that moves it to `OpenTESArena/CMakeLists.txt`.

OpenTESArena/CMakeLists.txt
```diff
@@ -78,6 +78,10 @@ IF (WIN32)
     ADD_DEFINITIONS("-D_SCL_SECURE_NO_WARNINGS=1")
 ENDIF()
 
+# Copy over required files
+FILE(COPY ${CMAKE_SOURCE_DIR}/data DESTINATION ${CMAKE_BINARY_DIR})
```
afritz1: The destinations should be `${CMAKE_CURRENT_BINARY_DIR}`, not `${CMAKE_BINARY_DIR}`. Sorry for not making that clear in my comment. The current binary directory ties it to where the current `CMakeLists.txt` outputs files.

Thunderforge: I'm guessing we want `${CMAKE_CURRENT_SOURCE_DIR}` too?

afritz1: The source directory refers to the top-level `CMakeLists.txt`, so it's already pointing to the right place.

Thunderforge: Gotcha. I just found out that `${CMAKE_CURRENT_SOURCE_DIR}` points to the subdirectory, so definitely not what we want. New commit is submitted.

PR #79, opened 2017-07-10
Adding settings folder for CLion to gitignore

Thunderforge:


afritz1:
Would you mind making the comment above it more specific for what IDE it's for? We can see it's for CLion here but I wouldn't normally know what an `.idea` file is. I'm just stating this because there are existing sections for Visual C++.

Thunderforge:
Oh, didn't realize that Visual C++ was an IDE. Do you just want me to have one IDE section or would you prefer individual ones for each IDE?

afritz1:
Maybe have named sections now so that they're easier to differentiate later.

PR #80, opened 2017-07-12
Preventing the app from crashing if there are no Arena files

Thunderforge:
It will now be handled the same way that missing options are handled.

### DIFF
--- a/OpenTESArena/src/Game/Game.cpp
+++ b/OpenTESArena/src/Game/Game.cpp
@@ -14,17 +14,15 @@
 #include "../Assets/CityDataFile.h"
 #include "../Assets/TextAssets.h"
 #include "../Interface/Panel.h"
-#include "../Math/Vector2.h"
-#include "../Media/AudioManager.h"
 #include "../Media/FontManager.h"
 #include "../Media/MusicFile.h"
 #include "../Media/MusicName.h"
 #include "../Media/PPMFile.h"
 #include "../Media/TextureManager.h"
-#include "../Media/TextureName.h"
 #include "../Rendering/Renderer.h"
 #include "../Rendering/Surface.h"
 #include "../Utilities/Debug.h"
+#include "../Utilities/File.h"
 
 #include "components/vfs/manager.hpp"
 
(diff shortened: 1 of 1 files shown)


OpenTESArena/src/Game/Game.cpp
```diff
@@ -36,7 +34,8 @@ Game::Game()
 	this->options = OptionsParser::parse();
 
 	// Initialize virtual file system using the Arena path in the options file.
-	VFS::Manager::get().initialize(std::string(this->options->getArenaPath()));
+    std::string arena = File::toString(this->options->getArenaPath());
```
afritz1: Um... I'm not sure this is doing what you want it to do. `File::toString()` takes a filename and returns all of the content of the file. We don't want to initialize the virtual file system with the content of the `ARENA` filename because it's a folder.

Thunderforge: Ah, didn't realize that. I guess I was thinking of [Java's File.toString()](http://docs.oracle.com/javase/8/docs/api/java/io/File.html#toString--), which returns the file path. Really what I'm wanting to do is take advantage of the check to see if the file actually exists.

afritz1: Maybe I want to add a `File::exists()` method that will check if the `ARENA` folder exists.

Thunderforge: I was thinking of doing the same thing, but upon closer inspection, I was a little unclear as to what the purpose of the `File` class was. Is it any type of file? Or a specific file for the engine?

afritz1: It's just a simple static class for common file operations. The only method right now is for reading all of a file's content into a string, but adding a method to check if a file/folder exists would also be good, so I'll do that.

Thunderforge: Great, were you going to do it on this branch or should I withdraw this PR?

afritz1: I'll do it on master, so you can delete this pull request.

PR #81, opened 2017-07-12
Adding SDL_GetBasePath() for macOS-friendly file loading

Thunderforge:
Per discussion in #74


afritz1:
I tested those changes on Windows, and while the path is kind of strange-looking with a mix of backslashes and forward slashes, it works. Maybe I could implement a method to clean up the slashes sometime.

Thunderforge:
Are the mix of back and forwards slashes actually causing issues? It was my understanding that Windows could handle both interchangeably.

afritz1:
It doesn't cause any issues other than cosmetic. I just like to have them be homogeneous.

PR #83, opened 2017-07-12
Noting that WildMIDI is optional

Thunderforge:


README.md
```diff
@@ -72,7 +72,7 @@ cd ..
 - [CMake](https://cmake.org/download/)
 - [OpenAL Soft 1.17.2](http://kcat.strangesoft.net/openal.html#download)
 - [SDL 2.0.4](https://www.libsdl.org/download-2.0.php)
-- [WildMIDI 0.4.0](https://github.com/Mindwerks/wildmidi/releases)
+- [WildMIDI 0.4.0](https://github.com/Mindwerks/wildmidi/releases) (Optional, no sound/music if not included)
```
afritz1: I think I would prefer if it said `(Optional, required for music)`. WildMIDI is only for music, so it doesn't affect sound. The sound code isn't implemented yet either, and it would require .VOC file decoding.

PR #85, opened 2017-07-13
Showing an error dialog if the application is going to crash

Thunderforge:
This is useful for macOS where std::getchar() is ignored when double-clicking on an app, so the app crashes immediately without warning.

### DIFF
--- a/OpenTESArena/src/Utilities/Debug.cpp
+++ b/OpenTESArena/src/Utilities/Debug.cpp
@@ -2,6 +2,7 @@
 #include <iostream>
 #include <map>
 #include <vector>
+#include <SDL_messagebox.h>
 
 #include "Debug.h"
 
@@ -63,7 +64,7 @@ void Debug::crash(const char *__file__, int lineNumber, const std::string &messa
 {
 	Debug::write(Debug::MessageType::Error, Debug::getShorterPath(__file__),
 		lineNumber, message);
-	std::getchar();
+	SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, "Error", message.c_str(), NULL);
 	exit(EXIT_FAILURE);
 }
 
--- a/OpenTESArena/src/Utilities/Debug.cpp
+++ b/OpenTESArena/src/Utilities/Debug.cpp
@@ -66,7 +66,7 @@ void Debug::crash(const char *__file__, int lineNumber, const std::string &messa
 		lineNumber, message);
     const std::string platformName(SDL_GetPlatform());
     if (platformName == "Mac OS X") {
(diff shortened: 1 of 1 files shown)


afritz1:
There are pros and cons to having a message box appear, since I've been fine just looking at the console for errors. I've thought about this feature before, but since you mention `std::getchar()` does nothing on macOS, then that's a good enough reason to have a message box. Maybe it should be in `options.txt` in case people want to use the console for errors.

Thunderforge:
To clarify, it does work if you run the Unix executable from the terminal, but if you double-click on OpenTESArena.app, it is ignored, since the user doesn't have anywhere to see the console output or register the enter key to. I'm actually surprised this hasn't come up already with the Windows .exe files, since you would be in the same situation I would think.

I suppose an `options.txt` property would be fine if people really would want to switch back.

afritz1:
The Windows release has a console window because it's built with the "console" subsystem, so I guess the macOS ".app" is the special case right now. Maybe it could check if the SDL platform is `Mac OS X` and do a message box then? There would be no need for an option in `options.txt` then.

Thunderforge:
Sure, I can check the platform if that's what you prefer.

afritz1:
For the time being, yeah. It will always use a message box on Mac then, regardless of whether it's on a console or through an ".app".

abelsromero:
Sorry to pup up, but would it make sense to integrate some logging mechanism?
That way the errors can be stored and retrieved independently of the platform. Also, the errors can be seen even after the console is closed.
The only problem is that it raises the issue of where to store the log files.

Thunderforge:
@abelsromero I agree that we should log out output to a file, but that seems like an additional measure. I've created #86 to talk about that. Mostly for this, I was concerned about Mac users wanting to know that something went wrong instead of having the app unexpectedly close.

OpenTESArena/src/Utilities/Debug.cpp
```diff
@@ -49,25 +50,30 @@ void Debug::write(Debug::MessageType type, const std::string &filePath,
...
 }
 
 void Debug::crash(const char *__file__, int lineNumber, const std::string &message)
 {
 	Debug::write(Debug::MessageType::Error, Debug::getShorterPath(__file__),
 		lineNumber, message);
-	std::getchar();
+    const std::string platformName(SDL_GetPlatform());
+    if (platformName == "Mac OS X") {
+        SDL_ShowSimpleMessageBox(SDL_MESSAGEBOX_ERROR, "Error", message.c_str(), NULL);
```
afritz1: In C++11, the desired way to indicate a null pointer is with `nullptr`, not `NULL`. The rest looks good, though.

abelsromero:
> Mostly for this, I was concerned about Mac users wanting to know that something went wrong instead of having the app unexpectedly close.

Agree, it's a nice quick win!

PR #87, opened 2017-07-15
Added MYS2 setup manual for Windows

abelsromero:
It took me more time than expected, but here is the manual in Markdown.

Some notes:
* I started sections at level2 indentation because I assume is it required so the file is included in the README.
* I had to use manual formatting for admonitions
* I wanted to call the the file `building_under_windows_msys2`, so that future manuals all share the smae prefix, but the `build` prefix is added to the .ignore so no way :/ I just called `setup_*`

cheers,


afritz1:
Great! Looks good. I hope some users will find it helpful. I'll add it to the Readme soon.

PR #88, opened 2017-09-19
Fix failing Travis CI build

Allofich:
Your Travis CI builds have been failing for a while due to lacking some dependencies. Adding them in fixed it. Looks like there aren't any other errors or warnings that crept in while the builds were failing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,7 @@ compiler: gcc
 group: stable
 
 before_install:
-  - sudo apt-get install libsdl2-dev libopenal-dev
-  - sudo apt-get install wildmidi
+  - sudo apt-get install libegl1-mesa-dev libgles2-mesa-dev libsdl2-dev libopenal-dev wildmidi
 
 install:
   - wget https://github.com/Mindwerks/wildmidi/archive/wildmidi-0.4.0.tar.gz


Allofich:
When I look at your commit history, I see a couple builds from a few months back that still say they are "pending," although if I click on it to open them it says they are done. It's probably nothing but it might be a good idea to cancel those if you can, just in case they are still held up in a queue somewhere.

afritz1:
Thanks for looking into that. I knew the builds were failing and I asked on the IRC chat for any ideas, but I didn't figure out why the libsdl2-dev library wasn't getting those dependencies anymore. There were a couple blog posts on the Travis website about them changing environments, but it didn't look like it would affect this particular project, so I didn't know what the source of the problem was then (maybe Trusty handling its SDL 2.0.6 dependencies differently from previous revisions?).

I don't think OpenMW had this problem. I see that [their Travis file](https://github.com/OpenMW/openmw/blob/e7e875081622b2f90f53d0c89456055949664bd5/.travis.yml#L35) includes libsdl2-dev, but maybe one of the other libraries there includes the ones that we were missing? I don't think it would've been OSG or MyGUI because those appear later in the list.

I also knew about the pending builds, but it also looked to me like they were done, so I can try investigating them again.

afritz1:
I hit the restart button for those two pending builds and they are now "failing", which is fine.

PR #99, opened 2018-01-15
Add support for building to macOS

Thunderforge:
I couldn't get WildMidi support working, but this got me to the point where I could build a fully-functional macOS app.

Addresses #74.


OpenTESArena/CMakeLists.txt
```diff
@@ -73,16 +75,38 @@ SET(TES_SOURCES
     ${TES_WORLD}
     ${TES_MAIN})
 
+SET(TES_DATA_FOLDER ${CMAKE_SOURCE_DIR}/data)
+SET(TES_OPTIONS_FOLDER ${CMAKE_SOURCE_DIR}/options)
+
 IF (WIN32)
     LIST(APPEND TES_SOURCES ${TES_RESOURCES})
     ADD_DEFINITIONS("-D_SCL_SECURE_NO_WARNINGS=1")
-ENDIF()
```
afritz1: I think there should be a small revision here, otherwise it is leaving out Linux.

```CMake
IF (WIN32)
    LIST(APPEND TES_SOURCES $(TES_RESOURCES))
    ADD_DEFINITIONS("-D_SCL_SECURE_NO_WARNINGS=1")
ENDIF()

IF (NOT APPLE)
    # Copy over required files...
    ...

    # Add the rest
    ...
ELSE (APPLE)
    # Info.plist properties
    ...
    ...
   # Add the rest
    ...
ENDIF()
```

Thunderforge: Good call. I've made the change.

afritz1: Oh, sorry, I meant to have those two lines at the top be their own if block, because with this change it will incorrectly set the Windows resources folder on Linux as well.

Thunderforge: I think I understood this time. Take a look and let me know.

afritz1: Yes, that looks more correct now. On another note, why were the curly braces around TES_RESOURCES changed to parentheses at `LIST(APPEND TES_SOURCES $(TES_RESOURCES))`?

Thunderforge: Looks like that was an oversight. Probably deleted it, then retyped it later and didn't catch it because it was under the Windows code. I've reverted it.

afritz1:
Okay, looks like I can merge it now. I'll make a note that music is not currently supported on the macOS version, although it probably wouldn't be too hard to get working if we got another macOS user in here to help look into it.

afritz1:
Also, I'm assuming this means we can show macOS support on the Readme now.

Thunderforge:
@afritz1 Yeah, I figure that we can do that. Do you want me to write it? Also, we had previously discussed me building a Mac release. How do we want to go about doing that?

afritz1:
I can add a short note in the introduction, and you can add some installation instructions if you want.

Since I haven't heard anyone else asking for a macOS release yet, I don't think it's in very high demand, _but_ you can do it if you like. I don't think 0.7.0 will happen until April-May.

No need to be formal about a release at this point -- just build the .app with the `data` and `options` folders in it, then give me the download link (Dropbox, Discord, etc., your choice) and I'll put it on the 0.6.0 release. I think macOS users usually have folders compressed in a .dmg format?

Thunderforge:
> I can add a short note in the introduction, and you can add some installation instructions if you want.

@afritz1 That sounds good. 

> I think macOS users usually have folders compressed in a .dmg format?

Pretty much. Technically, it's not really a folder; Mac `.dmg` files are comparable to ISO files in that they are a file that you "mount". It's pretty common to just stick the app, and a shortcut to the Applications folder on it, then the user drags the app to the Applications folder to "install" it. For instance, here's OpenMW's no-frills version once it's mounted.

macOS OpenMW 0.43 disk image

It's late in my time zone, so I'll take a look another day about sticking the app that I made onto a DMG and getting it to you, then updating the readme. Also, I started looking at Travis CI.

afritz1:
@Thunderforge, sure, we can get the macOS build and Travis CI working when it's best for you. No rush at all.

Issue #93, opened 2017-12-31
Clang warnings and adding macOS to Travis CI


Allofich:
I've noticed that the project is keeping very clean compile logs in Travis-CI GCC builds, and also when building in Visual Studio. I decided to test running it with clang enabled, and quite a few warnings popped up.

https://travis-ci.org/Allofich/OpenTESArena/jobs/323518887

It may be worthwhile to double as a light review of your code, maybe you'll see some other things you'd like to fix up while addressing these.

If you want to add clang as a compiler from here on, you can simply add it to the "compiler" option in the travis.yml file.

I also looked at adding OS-X builds, but it looks like a little work would be required to figure out how to get the dependencies correctly. Just adding OS-X to the travis.yml file, it will fail due to OS-X lacking the apt-get command.

On https://docs.travis-ci.com/user/installing-dependencies/, it explains how to set this up by OS. The example is:
`install:
  - if [ $TRAVIS_OS_NAME = linux ]; then sudo apt-get install foo; else brew install bar; fi`

and I was able to get past the error using this. But then, it still failed to download the dependencies. so we'd need to set it up to get appropriate OS-X dependencies. If it was all set up, in addition to adding another layer of checks on the code, it would be a step toward closing https://github.com/afritz1/OpenTESArena/issues/74.

afritz1:
Yeah, having it compile on both GCC and Clang would be objectively better I think.

I looked through that Travis CI log and all of the warnings are relatively benign, but I'll fix them anyway. I guess Clang is just a little more strict about some things like explicitly overriding pure virtual functions with other pure virtual functions.

We can get Travis CI to compile for macOS at some point, too.

Allofich:
Looks like all the Clang warnings are gone!

As for macOS/OS-X, is there anyone using or familiar with it? @Thunderforge, if you have the time, do think you could get it running on Travis CI?

Thunderforge:
@Allofich I am using macOS, but I've never tried to set it up. I'm up for the challenge though.

However, I would think we would need to get #74 finished first. I got stuck with WildMidi support (I'm a C++ newbie and am used to Java dependencies). Would you be okay with a PR that built a Mac app, but had no sound as a known issue?

afritz1:
From the looks of it, you already got it working on macOS (just without the music), so it seems to be almost supported. If you want, you can do a pull request for .travis.yml, and we can work from there. It's your choice (you're the residing macOS guy 😃).

[...]

PR #102, opened 2018-01-19
Adding macOS to Travis CI

Thunderforge:
Addresses #93.

Regarding multiple operating systems, I decided to follow [the way that OpenMW does it](https://github.com/OpenMW/openmw/blob/master/.travis.yml) by offloading the instructions to separate files, rather than the way [from the Travis CI documentation](https://docs.travis-ci.com/user/installing-dependencies/#Installing-Dependencies-on-Multiple-Operating-Systems), which will result in a messier build file.

All Mac and Linux builds passed when running against my local branch.


Thunderforge:
By the way, the Mac builds use [Homebrew](https://brew.sh/), which is equivalent to `apt-get`. This does not require `sudo` access. The Travis documentation says that turning off `sudo` [makes the Trusty builds go considerably faster](https://docs-staging.travis-ci.com/user/trusty-ci-environment/#Container-based-with-sudo%3A-false). I know that Linux build time isn't our bottleneck right now, but if we can run Linux without `sudo`, I think that it would be good.

afritz1:
Great! This gives us coverage on macOS compilers now. Putting the scripts in a `CI` folder was a good idea, too.

I think we left out `sudo` early on, but it was required for something, so we had to add it. I don't remember what it was, though.

PR #105, opened 2018-01-21
Fixing the icon not appearing on the macOS app

Thunderforge:
**Before:**
Generic app icon

**After:**
Custom app icon

Note that at some point, we should look into getting the app to be properly named "OpenTESArena", but that seems to be due to the CMake script calling the project "TESArena".


afritz1:
Yeah, it's a known issue that the executable isn't being named right, and I think it's because of ambiguous folder names on Linux or something, or that you can't have two things named OpenTESArena in the CMake file. I'm not sure. On Windows, I've just been hitting F2 and renaming the executable.

Thunderforge:
@afritz1 By the way, I put a link in IRC to a 0.6.0 Mac distribution.

Issue #107, opened 2018-01-24
No music on macOS


Thunderforge:
Sound isn't working in the 0.6.0 release of OpenTESArena. It should be fairly easy to get WildMidi set up with that.

afritz1:
Commit 43550ab1a4359025bce1f3fa2693343ea9c50f70 might be relevant. The `MidiConfig` string wasn't getting the base path prepended, so it likely wasn't looking inside the macOS .app correctly before.

afritz1:
I think this can be closed.

PR #115, opened 2018-03-24
Simplify Travis-CI files

Allofich:
This is intended to clean up and simplify the Travis-CI files, particularly for OS-X builds, by making the commands as minimal as I could make them. WildMIDI is downloaded and built for OS-X now, but not yet installed and used in the compilation of OpenTESArena.

This is the result of a lot of trial-and-error. I don't use OS-X or Linux, so the criteria for me was whether or not Travis-CI successfully built. Input from others who are more knowledgeable is welcome.

Changes:
`travis.yml`
**Removed** `group: stable` - Not necessary
**Changed OS-X to a single Clang build** - As far as I can see, master is currently building both of these targets with Clang, not one with Clang and the other with GCC. They both are reporting

```
-- The C compiler identification is AppleClang 8.0.0.8000038
-- The CXX compiler identification is AppleClang 8.0.0.8000038
```

I experimented with trying to get one build to use GCC, but couldn't get it to work. I don't know that it's actually worthwhile to build with both compilers for OS-X; I have no idea what the usual compiler used on OS-X is. At any rate, the OS-X build can be simplified to just `os: osx`, which is nice.

**Removed** `osx_image: xcode8` - Not necessary, and if present causes the below warning to appear

```
Warning! PATH is not properly set up, '/Users/travis/.rvm/gems/ruby-2.0.0-p648/bin' is not at first place,
         usually this is caused by shell initialization files - check them for 'PATH=...' entries,
         it might also help to re-add RVM to your dotfiles: 'rvm get stable --auto-dotfiles',
         to fix temporarily in this shell session run: 'rvm use ruby-2.0.0-p648'.
```

(Note: `osx_image: xcode8.3` does not create the above warning)

`CI/before_install.linux.sh` 
**Removed** `wildmidi` **from the** `sudo apt-get install` **command** - not necessary

`CI/before_install.osx.sh`
**Removed all brew commands except for** `brew install sdl2` - other commands not necessary

I also looked into the below warnings in the OS-X builds but I think that these are problems on the Travis-CI end. They appear even with a practically empty travis.yml file.

```
Fix WWDRCA Certificate
Unable to delete certificate matching "0950B6CD3D2F37EA246A1AAA20DFAADBD6FE1F75"security: AppleWWDRCA.cer: already in /Library/Keychains/System.keychain
```

```
/Users/travis/.travis/job_stages: line 475: yarn: command not found
```

```
/Users/travis/.travis/job_stages: line 166: shell_session_update: command not found
```

Additionally, when WildMIDI is built in the OS-X builds, the below warning appears.

```
CMake Warning (dev):
  Policy CMP0042 is not set: MACOSX_RPATH is enabled by default.  Run "cmake
  --help-policy CMP0042" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.
  MACOSX_RPATH is not specified for the following targets:
   libwildmidi_dynamic
This warning is for project developers.  Use -Wno-dev to suppress it.
```

but it also appears for the Travis-CI build logs at https://github.com/Mindwerks/wildmidi/tree/master/CI.

I could not get the built WildMIDI to be used in the compilation of OpenTESArena, because `make install` will fail for the OS-X build. If we can get this I suppose that we'll have MIDI support on OS-X.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: cpp
 sudo: required
-group: stable
 
 matrix:
   include:
@@ -24,11 +23,9 @@ matrix:
       env:
         - MATRIX_EVAL="export CC=clang-3.6 && export CXX=clang++-3.6"
     - os: osx
-      osx_image: xcode8
       env:
         - MATRIX_EVAL="export CC=gcc && export CXX=g++"
     - os: osx
-      osx_image: xcode8
       env:
         - MATRIX_EVAL="export CC=clang && export CXX=clang++"
 
--- a/CI/before_install.linux.sh
+++ b/CI/before_install.linux.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-sudo apt-get install libegl1-mesa-dev libgles2-mesa-dev libsdl2-dev libopenal-dev wildmidi
+sudo apt-get install libegl1-mesa-dev libgles2-mesa-dev libsdl2-dev libopenal-dev
--- a/CI/before_install.osx.sh
+++ b/CI/before_install.osx.sh
@@ -1,12 +1,3 @@
 #!/bin/bash
 
-brew update
-
-#Upgrade packages included on the default image
-#See https://docs.travis-ci.com/user/reference/osx/#Compilers-and-Build-toolchain
-brew outdated cmake || brew upgrade cmake
-brew outdated pkgconfig || brew upgrade pkgconfig
-
-#Install packages not included on the default image
-brew install openal-soft
 brew install sdl2
--- a/CI/install.osx.sh
+++ b/CI/install.osx.sh
@@ -1,1 +1,5 @@
 #!/bin/bash
+
+wget https://github.com/Mindwerks/wildmidi/archive/wildmidi-0.4.0.tar.gz
+tar -xzvf wildmidi-0.4.0.tar.gz
+pushd wildmidi-wildmidi-0.4.0 && mkdir build && pushd build && cmake .. && make && popd && popd


afritz1:
It looks like you were able to safely remove the unnecessary commands from the `before_install` scripts. I think @Thunderforge just added some of those extra `brew` commands for redundancy (but I guess Travis CI updates them already).

It looks like the macOS build will default to GCC if you don't specify any build options. AppleClang is the Xcode front-end (I guess?), but it actually does differentiate between GCC and Clang behind the scenes as you can see by the "Check for working C/CXX compiler" message below it. I think we might want to add back part of the macOS code in `.travis.yml`:
```yaml
- os: osx
  env:
    - MATRIX_EVAL="export CC=clang && export CXX=clang++"
```

macOS might be using different implementations of standard libraries behind the scenes (i.e., LLVM libc++ instead of libstdc++), which may cause compiler-specific warnings/errors, so it might still be beneficial to have full coverage on macOS (i.e., both GCC and Clang builds). However, I don't see us actually *needing* GCC on macOS, so maybe we don't need that second build after all.

I don't know about the problems with WildMIDI, but some reading shows that the CMake policy warning is a requirement from newer versions of CMake, and should be set by WildMIDI's CMake script. Something like:
```CMake
if (APPLE)
    set(CMAKE_MACOSX_RPATH ON)
endif()
```

Allofich:
> It looks like the macOS build will default to GCC if you don't specify any build options. AppleClang is the Xcode front-end (I guess?), but it actually does differentiate between GCC and Clang behind the scenes as you can see by the "Check for working C/CXX compiler" message below it. I think we might want to add back part of the macOS code in .travis.yml:

It's in now.

> I don't know about the problems with WildMIDI, but some reading shows that the CMake policy warning is a requirement from newer versions of CMake, and should be set by WildMIDI's CMake script. 

Yeah I got the same idea. It's probably an issue on WildMIDI's end.

Allofich:
If we could just get WildMIDI to be used in the compilation of openTESArena, we could maybe close https://github.com/afritz1/OpenTESArena/issues/107 (since Travis-CI building successfully with WildMIDI on OS-X would demonstrate it can be compiled with MIDI support on OS-X).

I'm a complete beginner at OS-X stuff, it's all experimentation via Travis-CI. I have tried doing the same thing as for the linux builds by using `make install` (no `sudo` needed) for the OS-X build after running `make`, but I get the error `make: *** No rule to make target `install'.  Stop.`

I tried to look up this error and found suggestions about 1) Not being in the right directory (I think we have the right directory, so not this) and 2) Not all dependencies installed (does WildMIDI need some dependencies not present in the OS-X environment?) I'm not sure that either of these suggestions apply to this situation.

Thunderforge, since you are actually a Mac user, maybe there is something obvious to you here?

afritz1:
I made a pull request to fix the WildMIDI warning and they merged it, but it turned out to be a very mundane issue about CMake 3.0+ using "rpaths" on macOS by default and it requiring CMP0042 to be set in the CMakeLists.txt. I think getting WildMIDI to build is entirely on our side, and just something we need to figure out.

Thunderforge:
@afritz1 @Allofich Regarding WildMidi on a Mac, I am not entirely sure why I wasn’t able to link it up. I only started using C++ a few months ago and it’s possible there is something obvious I’m missing. All of the other dependencies are available via `brew` (Mac equivalent to `apt-get`), but [WildMidi is not](https://github.com/Mindwerks/wildmidi/issues/172).

afritz1:
The reason why I said I thought it would be a simple fix is because the .app probably just needs to have the libwildmidi_dynamic.dylib (or whatever it's called) dropped in there as part of the build process. If it was crashing for you on start-up with that .dylib error then that's probably it. It'd be a similar error on Windows (i.e., "Your computer is missing SDL2.dll" or something like that for SDL).

Issue #69, opened 2017-04-28
Game crashes when running with MIDI


Allofich:
If I run the game with MIDI active, it crashes when I click "Start New Game."

Specifically, if I have eawpats.zip extracted into the Data folder, the game starts and plays MIDI, but crashes when I click Start New Game. If I don't have eawpats in the Data folder, the game doesn't crash.

Allofich:
The game also crashes on exit when I'm using eawpats.

afritz1:
You're on Windows, right? I remember having similar issues. It might be caused by using the wrong build of `wildmidi_dynamic.dll`. I think I fixed it by building the debug version of WildMIDI and using the .dll that came from that. @psi29a, any other advice?

Allofich:
Yes, I'm on Windows 7, 64-bit. I'll try doing the same thing you did.

Allofich:
Yes, that worked. And then I built the Release build of wildmidi and used that, and the crash returned. So, back to the Debug build.

psi29a:
Nice find. You building with MSVC? Have you tried using our binaries from WildMIDI 0.4.1?

afritz1:
Yeah, I'm using Visual Studio 2015. I set up CMake to build OpenTESArena with the wildmidi-0.4.1-win64/msvc folder just now, and it runs fine in Visual Studio, but if I run the executable itself with `wildmidi_dynamic.dll` in the directory, then it crashes after clicking "Start New Game", just like with Allofich's crash. 

If I build and use the debug version of `wildmidi_dynamic.dll` with the OpenTESArena executable, it doesn't crash.

Allofich:
I also use MSVC 2015. From the binaries, do you mean wildmidi.exe and wildmidi-static.exe in the msvc folder of WildMIDI 0.4.1? I tried it, and at first it didn't start because of not finding wildmidi.cfg. I copied that into the folder and tried it on a midi, and while the program ran, I got no sound. It was a Daggerfall midi and maybe it's not supported? And, it crashes when I exit the program. So, either I'm not using it right or there are some problems with it.

afritz1:
[...]

PR #122, opened 2018-05-16
Make Travis CI builds use latest WildMIDI code

Allofich:
Since https://github.com/afritz1/OpenTESArena/issues/69 is fixed by new changes in WildMIDI, and another OpenTESArena-related issue (and a Daggerfall-related issue) are still open there and seem to be waiting for fixes, it seems to me that it's better to be testing with the latest code. On the off chance that some incompatibility happens between the latest OpenTESArena code and the latest WildMIDI code it would be good to know.

### DIFF
--- a/CI/install.linux.sh
+++ b/CI/install.linux.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-wget https://github.com/Mindwerks/wildmidi/archive/wildmidi-0.4.0.tar.gz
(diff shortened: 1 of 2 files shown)


afritz1:
We can do this for Travis CI, and I probably also want to add a note to the installation section for Visual Studio users, since the WildMIDI 0.4.3 changes affect them the most.

I don't think we want to raise the WildMIDI requirement for the whole project though since some Linux distributions don't have the latest WildMIDI available.

PR #127, opened 2018-05-27
Improved weapon fitting in modern mode

enderdrag64:
I changed the offset of the weapons depending on which side of the screen they intersected with. This way I could preserve the original width without causing the weapons to cut off. There can be some improvements made to the offset values, I didn't measure them exactly to the pixel.

### DIFF
--- a/OpenTESArena/src/Assets/MIFFile.cpp
+++ b/OpenTESArena/src/Assets/MIFFile.cpp
@@ -183,7 +183,7 @@ int MIFFile::Level::load(const uint8_t *levelStart)
 	// size in WILD.MIF (six bytes short of where it should be, probably due to FLAT
 	// tag and size not being accounted for, causing this loader to incorrectly start
 	// a second level six bytes from the end of the file).
-	return static_cast<int>(std::distance(levelStart, tagStart));
+	return static_cast<int>(levelEnd - levelStart);
 }
 
 int MIFFile::Level::getHeight() const
--- a/OpenTESArena/src/Assets/MIFFile.cpp
+++ b/OpenTESArena/src/Assets/MIFFile.cpp
@@ -183,7 +183,7 @@ int MIFFile::Level::load(const uint8_t *levelStart)
 	// size in WILD.MIF (six bytes short of where it should be, probably due to FLAT
 	// tag and size not being accounted for, causing this loader to incorrectly start
 	// a second level six bytes from the end of the file).
-	return static_cast<int>(levelEnd - levelStart);
+	return static_cast<int>(std::distance(levelStart, tagStart));
 }
 
 int MIFFile::Level::getHeight() const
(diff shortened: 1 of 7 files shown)


afritz1:
Thanks for looking into this! While I appreciate your work, I'm not sure if I want to go in this direction with the changes. Reading over it, it just feels too specific of a solution that makes assumptions about Arena's weapon animations. I'm not sure if it can be generalized, so I've just gone ahead with the easy solution and made it so all weapon animations stretch to fit the window in modern mode in commit fabfe0084eb4bd3d0b8adc69d658111adbb13686.

`WeaponAnimation::isLeft()` and `isRight()` having different return conditions feels like a hack to me, and it looks like it assumes that the player only has right-handed weapons? If there was ever a left-handed weapon, then the code would break (not saying there will be; just saying that's what might happen).

I think a better solution to the weapon offset values (55 and 20) would be a formula that uses `Renderer::getLetterboxAspect()` as input. Also, a small nitpick but the letterbox mode should be obtained from the `Options` object and not the renderer. I intentionally made `Renderer::letterboxMode` a write-only variable (maybe I should comment that).

With regards to your changes in the software renderer, I think they are ultimately unnecessary (I know... I deliberate over these small things, too). The compiler should optimize that integer division into a right shift, so it might actually be slower with a float multiply. Besides, modern CPUs are so complex now that it'd probably run the exact same either way.

It's just the nature of the weapon animations that makes them not very flexible for other aspect ratios. If they were 3D models like in Battlespire, then I think we'd be able to do more with them (but that's not something I think I'm ever going to have mod support for).

Also, for future pull requests, try and remember to make your changes on a branch other than master so it's easier to delete, and try to keep unrelated changes in separate pull requests.

Overall, it was a good first pull request, and thanks for spending your time on it! I just felt that in this particular case a different solution was more fitting (no pun intended).

enderdrag64:
Thanks for the feedback! I’m sorry I didn’t realize Renderer::letterboxMode was write-only. I'd tried using Renderer::getLetterboxAspect(), but if I'd write say bool a = renderer.getLetterboxAspect() == 4/3; it's never be true. I never even thought to pull it from the options menu. 
For future pull requests I'll definitely put them in a different branch, I wasn't sure exactly how that worked. I'll also keep in mind your other critiques in the future.

afritz1:
Yeah, no worries. What I meant with the letterbox aspect is that it'd be something like:
```C++
int offset = <some formula with letterbox aspect>; // And isLeft()/isRight().
int weaponX = weaponOffset.x + offset;
```
The letterbox modes themselves are meant to be on the user-facing side for convenience, while the letterbox aspect is what the engine uses for all the related calculations.

I have functional programming on the mind with this codebase, so I try to convert assignments into expressions whenever I can :wink:. It makes code a lot more traceable when values are constant.

With branches, just make a new branch with a short name like `my-new-feature` and then make your commits to that. It's just a convenience thing and it helps when you want to keep unrelated work separate.

PR #128, opened 2018-05-31
Added Windows build instructions

AnotherFoxGuy:
Added build instructions on how to setup a build environment for building with vcpkg and Visual Studio


afritz1:
Nice, thanks. Looks good to me. Does vcpkg set environment variables for each of the packages automatically (`SDL2DIR`, `WILDMIDIDIR`, etc.)?

We can make small improvements to it in the future as needed.

PR #130, opened 2018-07-28
Update msys2 manual + minnor README improvements

abelsromero:
First, 
    keep up the good job here!!

Then, 
   A part from what the PR name stated I validated that the bug related to WildMidi does not affect the MSYS2 build, so I updated the manual to use the released version 0.4.2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,22 +100,24 @@ cd ..
 
 ## Building from source
 
-#### Project dependencies:
+### Project dependencies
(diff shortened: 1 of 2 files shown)


afritz1:
Looks good, thanks Abel! I hope to have more time to spend on this project soon.

And yeah, I believe the WildMIDI people and @Allofich came to the conclusion that it was a Visual Studio optimization bug, so MSYS2 wouldn't be affected.

PR #139, opened 2019-02-24
Started working on an OpenGL backend.

LordDeimos:
There are some issues with ceilings, raised platforms, and chasms, but for the most part the basics are done.
<details><summary>Some screenies for comparison (opengl on the right):</summary>
<p>

</p>
</details>
<br/>
There is still a long way to go before it behaves the same as the Software Renderer, but I think it does an ok job so far.


afritz1:
That's interesting that you're working on OpenGL stuff. I wasn't aware of any of this. I scanned through the changes and it looks like a lot of work was put into this (thank you for referencing the contributing guide, too). The screenshots look promising.

I wrote a lot of notes but here are my points:
1) My intention in the long run is to make the software renderer efficient enough that it doesn't matter whether there is a hardware-accelerated option.
2) The renderer is still in the design stage, so having two renderers to keep up to date will decrease the speed of development.
3) I would like to not have any graphics API dependencies (other than SDL), even if it's ifdef'd out in C++ and CMake.

I know people would like the option for a hardware renderer but I just don't think I'm going to want the extra overhead of maintaining it for the coming months/years. I _already_ spend enough time on the software renderer (and would like to spend more, actually), so that means maintaining a hardware renderer alongside that would be impractical. I don't want there to be a desync between software and hardware renderer features at any point because the gap could always increase and then things don't get updated in a timely manner.

There are other implications with bringing on OpenGL as a feature of the application. I would inevitably be asked to fix a range of OpenGL-based bugs that I'm just not interested in spending time on. I'm not planning on touching OpenGL at all, but by having it in the application, it would be inevitable. It's a slippery slope that may turn into another time sink because I would want it to be done 'well'.

I already sunk several months into an OpenCL ray tracer that I dumped in favor of the software renderer after version 0.2.0 and I'm not really interested in bringing on another graphics API right now.

My software renderer is not in a stable state yet and I might want to redesign parts of it in the not-too-distant future. If I do that, then the OpenGL code will need to be brought up to speed, which I don't plan on spending any time on.

For the foreseeable future, I would like to remain focused on the software renderer being the only option for the application. If I do ever bring back a graphics API, it would most likely be Vulkan. I don't like the idea of bringing on an ancient dependency like OpenGL because it has a lot of historical baggage and can cause headaches throughout development. Vulkan would presumably also take a lot of time so that's why I'm still leaning towards keeping a software renderer for everything.

I will probably need to sleep on it before I decide on pulling this into master.

LordDeimos:
I admit, a hardware renderer is a large can of worms to open. I guess the only consolation is you don't necessarily need to maintain both personally, I can (when I can) cover issues with this renderer so it isn't on your plate to worry about, that is the power of open source. OpenGL is also not the nicest thing to work with, and I have thought about either moving to or having a second render back-end in Vulkan (though having two is going to exacerbate issues I feel), it is more a proof of concept at the moment that seems to work all right for what it is. I'm happy to work on my fork until it reaches a more feature-complete point that meets with your approval, and even if it doesn't, it has been good to use the stuff I've been teaching myself into a more practical setting.

afritz1:
It's good to hear your thoughts on it too. I slept on it and my new thoughts are that, if we did go the OpenGL route, wouldn't we want to use OpenSceneGraph instead of raw OpenGL? That's kind of the can of worms I'm talking about. (and even then, I think I'd rather use VulkanSceneGraph, but I don't know the state of that right now)

I agree that I might not necessarily need to maintain the hardware renderer myself. It's just a matter of still being able to maintain the whole codebase myself, which would include the hardware renderer.

I want to focus all my effort on the software renderer so I can make it _good_, and having a hardware renderer on the side that may or may not stay well-maintained might just end up taking away brain power from other things.

So you can certainly keep the hardware renderer in your own fork, but it won't be in master yet. If it does ever get pulled into master, I won't be the one maintaining it.

afritz1:
Just wanted to say that I'm _really_ glad there isn't another renderer to maintain right now. The software renderer is still this big unfinished thing that takes up a lot of cognitive load.

Issue #143, opened 2019-03-06
C++17


afritz1:
It's 2019 and there are several places in the codebase that could immediately take advantage of C++17 features.
- `std::byte`
- `std::clamp`
- `std::filesystem`
- `std::optional`
- `std::string_view`
- `std::variant`
- Guaranteed copy elision
- Structured bindings
- Terse `static_assert`
- ...

I briefly tested it on Windows 10 by changing the `14` to `17` in `set(CMAKE_CXX_STANDARD 14)` in `CMakeLists.txt` and everything just worked in Visual Studio 2017 (that is, it correctly added the `/std:c++17` compiler setting).

I believe the current GCC requirement of the project is 5.4. That was the same as OpenMW for no particular reason.

According to https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B17_features, it looks like we need GCC 7 and Clang 4 for Travis CI. GCC 5.4 is from 2016 and GCC 7.1 is from 2017, I think?

Anything else I forgot regarding upgrading to C++17 should go here. Maybe I should create a `cpp17` branch with some features and then try to have Travis build that?

afritz1:
This is precisely a year after https://github.com/afritz1/OpenTESArena/issues/113, interesting. I think the compiler support should be here now.

Also, like I said in a comment in that issue, GCC 6.1 finally adds `std::hash` support for enum classes, so I wouldn't need to define pointless hash functions anymore.

afritz1:
Upgrade complete.

afritz1:
Can confirm it builds on GCC 7.3.0 on Ubuntu 18.04 64-bit.

PR #146, opened 2019-04-06
Upgrade to C++17.

afritz1:
Setting the C++ standard in CMake is hopefully enough for MSVC, GCC, and Clang. I'm more worried about Travis CI possibly needing to download and install g++-7 every time but we'll see.

References issue #143.


afritz1:
Looks good to me. There might be issues on a per-user basis but we can look at those later.

PR #147, opened 2019-04-07
Get Travis CI Clang working with more C++17 features

afritz1:
Clang works with some C++17 features like `std::string_view` but not `std::optional`, likely due to Travis CI using an older libstdc++. Probably need to get it to use libc++ instead.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ matrix:
     - os: osx
       osx_image: xcode9
       env:
-        - MATRIX_EVAL="brew install gcc && CC=gcc-7 && CXX=g++-7"
+        - MATRIX_EVAL="brew update-reset && brew install gcc && CC=gcc-8 && CXX=g++-8"
 
 before_install:
   - eval "${MATRIX_EVAL}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ matrix:
       env:
         - MATRIX_EVAL="export CC=gcc-7 && export CXX=g++-7"
     - os: osx
+      osx_image: xcode10.1
       env:
         - MATRIX_EVAL="export CC=clang && export CXX=clang++"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ matrix:
       env:
         - MATRIX_EVAL="export CC=gcc-7 && export CXX=g++-7"
     - os: osx
-      osx_image: xcode10.1
       env:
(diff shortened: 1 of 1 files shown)


afritz1:
Can't use `std::experimental` namespace because Clang's `std::experimental` doesn't have some member functions for `std::optional` like `has_value()`.

If push comes to shove, maybe will need to create an `Optional` class that wraps `std::optional` on MSVC and GCC, and wraps `std::unique_ptr` on Clang 🤷‍♂️

psi29a:
That was the problem I faced with C++17... I wanted to replace many of the boost::filesystem stuff with std::filesystem, but it was pushed to std::experimental::filesystem :(

psi29a:
Another possibility is to use gcc instead of clang/xcode om macos... have homebrew install it

afritz1:
Looks like Travis CI had an internal error or something with Homebrew:
```
/usr/local/Homebrew/Library/Homebrew/brew.rb:27: syntax error, unexpected ':', expecting keyword_end
      Warning: #{e.message}
              ^
/usr/local/Homebrew/Library/Homebrew/brew.rb:28: syntax error, unexpected '!', expecting keyword_end
The command "eval "${MATRIX_EVAL}"" failed and exited with 1 during .
Your build has been stopped.
```

Allofich:
After searching online and doing some trial-and-error, I found the `osx` part will build correctly if you replace 

`- MATRIX_EVAL="brew install gcc && CC=gcc-7 && CXX=g++-7"`

with

`- MATRIX_EVAL="brew update-reset && brew install gcc && CC=gcc-8 && CXX=g++-8"`

The `brew update-reset` part fixes the syntax error. But it will then complain:

```
CMake Error at /usr/local/Cellar/cmake/3.9.2/share/cmake/Modules/CMakeDetermineCCompiler.cmake:48 (message):
  Could not find compiler set in environment variable CC:
  gcc-7.
Call Stack (most recent call first):
  CMakeLists.txt:2 (PROJECT)
CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
```

Switching to 8 will fix it. You still need the `brew update-reset` part or it will fail with the syntax error again.

afritz1:
Looks like it works. Thanks @Allofich! I'm not sure what the official GCC requirement of the project is now but I assume it's at least 7.1.

psi29a:
Sweet, nice work. Might have to use this in OpenMW when the time comes.

I just pushed OpenMW to C++14 ;)

PR #150, opened 2019-05-12
Fix OS X build on Travis CI

Allofich:
For some reason the OS X builds have stopped working, with the message:

```
CMake Error at /usr/local/Cellar/cmake/3.9.2/share/cmake/Modules/CMakeDetermineCCompiler.cmake:48 (message):
  Could not find compiler set in environment variable CC:
  gcc-8.
```

After trial-and-error, I found that by updating the image to 10.2, the latest one here:
https://docs.travis-ci.com/user/reference/osx/
it is fixed. While experimenting I got a message that the gcc version installed on this image is 8.3, and it seems to build okay.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,7 @@ matrix:
       env:
         - MATRIX_EVAL="export CC=gcc-7 && export CXX=g++-7"
     - os: osx
-      osx_image: xcode9
-      env:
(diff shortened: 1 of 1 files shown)


afritz1:
It's interesting that you removed the GCC parts in the script and it still uses GCC by default. I'm confused, but if it works it works. Thank you.

PR #151, opened 2019-05-31
📚 Update Windows build instructions for VS 2019 

AnotherFoxGuy:
And CMake 3.14


afritz1:
:+1:

I don't see any CMake changes so I assume that's automatic.

PR #152, opened 2019-06-01
📦 Added PMM for managing VCPKG packages

AnotherFoxGuy:
Added PMM (https://github.com/AnotherFoxGuy/pmm) for managing VCPKG on windows.
PMM takes care of downloading, bootstrapping and installing the dependencies so you don't have to run the commands manually.
It also ensures that the deps versions are always the same.

PMM can be disabled by setting `USE_PACKAGE_MANAGER=OFF`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,20 @@ ENDIF (COMMAND cmake_policy)
 PROJECT(OpenTESArena)
 
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/")
+include(CMakeDependentOption)
+
+cmake_dependent_option(USE_PACKAGE_MANAGER "Use VCPKG for managing packages" ON MSVC OFF)
+
+IF (USE_PACKAGE_MANAGER)
+    include(pmm)
+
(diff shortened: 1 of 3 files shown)


afritz1:
I know package managers are convenient but in all honesty this project barely has any dependencies. Is it really that much of a barrier to download SDL2 and OpenAL Soft and set the environment variables? It's like vcpkg and pmm are doubling the amount of dependency fluff in the repo.

I'm not planning on adding new dependencies, so SDL2 and OpenAL Soft (and optional WildMIDI) will probably be the upper limit for the project. Maybe in the distant future when I want to add a Vulkan renderer then a package manager will look more enticing.

If we do include this, it needs to be something that developers opt into. This line seems to set it on by default for MSVC? How does it behave on Linux/macOS? Is it off then? Maybe it would be more obvious if it was inside some `if (MSVC)` condition or something.
```CMake
cmake_dependent_option(USE_PACKAGE_MANAGER "Use VCPKG for managing packages" ON MSVC OFF)
```
Having pmm point to your personal website seems non-ideal to me (the lifetime aspect of it). If I do merge this commit, then it tells me you're going to maintain everything involving vcpkg and pmm, and if you don't feel like working on the project anymore, then you'll remove them.

afritz1:
Going to pass on this for now. Maybe check again in a year or something.

PR #158, opened 2019-08-08
[Don't merge] Add exclusive fullscreen mode

afritz1:
Making a pull request for this since I haven't figured out why Windows 10 freaks out when switching from exclusive fullscreen to borderless fullscreen.

I'd like exclusive fullscreen to only be the desktop resolution for now but I'm having trouble getting SDL to do that with `SDL_SetWindowFullscreen()` and `SDL_SetWindowMode()`.

There are also some DPI problems that crop up if the fullscreen resolution is different from the desktop, so that's another reason to match the desktop resolution.


afritz1:
Also probably want to change ScreenWidth/Height to WindowWidth/Height so it's clear it only applies to windowed mode, not borderless or fullscreen.

afritz1:
Closing this to attempt it in a new PR.

PR #168, opened 2019-12-19
[Don't Merge] Added the start of some physics bounding volumes

faha223:
Added a Collider3D class and some Bounding volumes (capsule, axis aligned bounding box, quad, and cylinder) that inherit from it. Once these are all fleshed out they'll be able to determine if they're colliding with one another in 3D space. This will help to implement a better, more complete movement system for the game.


afritz1:
Going to start seriously investigating a collision system next release and will likely solve it in a different way than how this PR does.

Issue #166, opened 2019-12-14
Implement 3D positional audio


afritz1:
This is a feature not present in the original engine. Essentially the OpenAL Soft code needs to be modified so it can play both 1) sound on camera, and 2) sound at a 3D point.

The audio manager should send position/rotation values of the camera, etc. to OpenAL Soft in its update() method every frame.

Places of note:
- [AudioManagerImpl::playSound()](https://github.com/afritz1/OpenTESArena/blob/cf4198c1d1bd276253df44839d5d53004d9742d2/OpenTESArena/src/Media/AudioManager.cpp#L525)
- [AudioManagerImpl::update()](https://github.com/afritz1/OpenTESArena/blob/cf4198c1d1bd276253df44839d5d53004d9742d2/OpenTESArena/src/Media/AudioManager.cpp#L669)

afritz1:
Merged in #170.

PR #170, opened 2019-12-28
[Don't Merge] Implementation of 3d positional audio

Plettro:
I implemented a simple 3d audio system that satisfies the requests stated in issue #166 . Here are the changes:

- Added a couple of methods to the _AudioManager_ class that make it possible to change the global listener's position and orientation
- Modified the _playSound_ method to optionally accept a second argument that specifies the position of the source (if left unspecified, the sound will be played "everywhere")
- Updated the part of the code that make use of said features accordingly.

This is my first time using git(hub), so I apologize in advance if I messed up something.

Also, sorry for all of the unnecessary space removals, it was my editor's fault and I don't really know how to reverse them...

Let me know if you want me to change anything or if you think my approach could be improved :)

### DIFF
--- a/OpenTESArena/src/Entities/Player.cpp
+++ b/OpenTESArena/src/Entities/Player.cpp
@@ -295,7 +295,7 @@ void Player::handleCollision(const WorldData &worldData, double dt)
 			}();
 
 			// -- Temporary hack for "on voxel enter" transitions --
-			// - @todo: replace with "on would enter voxel" event and near facing check.			
+			// - @todo: replace with "on would enter voxel" event and near facing check.
 			const bool isLevelUpDown = [&voxelData]()
 			{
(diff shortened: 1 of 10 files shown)


afritz1:
Thank you. It's cool to finally have doors playing 3D sounds 🎉 

This PR is pretty much what I was looking for. It's smaller than the other active ones so I should be able to merge it fairly easily. Will just go through and make some small revisions and experiment with a ListenerData struct passed to `AudioManager::update()` instead of setting listener position/orientation directly.

I'll probably also add an option to choose between 2D and 3D world sounds because some people might prefer the authentic (mildly bad) Arena experience where they can't tell where a fireball is coming from.

afritz1:
Seems ready to merge into master, thanks again.

PR #184, opened 2020-07-21
Texture manager refactor

afritz1:
Re-designed the texture manager so it can fail to look-up a texture. Added native support for storing/handling 8-bit images and started using a new ID-based system for fast texture look-up.

Ran into some issues with dangling texture references if texture IDs are obtained after a texture reference look-up because the internal texture buffer in the texture manager might get resized, so just need to be careful by getting all texture IDs ahead of time. At least the texture look-up functions are const now.

### DIFF
--- a/OpenTESArena/src/Game/Game.cpp
+++ b/OpenTESArena/src/Game/Game.cpp
@@ -74,9 +74,6 @@ Game::Game()
 		static_cast<Renderer::WindowMode>(this->options.getGraphics_WindowMode()),
 		this->options.getGraphics_LetterboxMode());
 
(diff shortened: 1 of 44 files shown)


afritz1:
Will be a little more important now to run sanitizers on the codebase before releases because of possible bugs like this but it's a net gain in the long run. I'm also a little confused how my old texture manager design never led to this same dangling texture reference bug. My best guess is that the `std::unordered_map` buckets had a small enough intersection with previously-obtained texture references in a certain scope that none of the buckets got re-allocated and caused bugs.

PR #186, opened 2020-08-06
Entity animation refactor

afritz1:
Changed from a slightly-monolithic EntityAnimationData pattern to a shared + unique pattern with EntityAnimationDefinition and EntityAnimationInstance.

Realized during this that the EntityManager shouldn't be returning Entity pointers at all because Entity::setPosition() can cause their chunk coordinate to change which could make the entity instance move around in memory and cause the entity pointer to become dangling. So in the near future: add some EntityRef wrapper like the texture manager.


afritz1:
Did a little bit of testing with an address sanitizer turned on in GCC but it didn't catch the problem with Entity::setPosition() being before the animInst updating, so that's a little unsettling. Anyway, merging this in now and should do the entity ref stuff soon.

PR #193, opened 2021-01-05
Refactor OptionsPanel render() into smaller methods

Thunderforge:
I was interested in making some changes and found the `render()` method a bit difficult to follow. Breaking into smaller methods for clarity.

### DIFF
--- a/OpenTESArena/src/Interface/OptionsPanel.cpp
+++ b/OpenTESArena/src/Interface/OptionsPanel.cpp
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <limits>
+#include <optional>
 
 #include "SDL.h"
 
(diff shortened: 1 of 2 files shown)


afritz1:
Good idea. The panel files are in various stages of revision and splitting up render() into smaller parts is reasonable. I will switch to this branch and quickly do some formatting fixes to match the rest of the codebase.

Thunderforge:
Thanks for cleaning up the formatting. Is there a style guide you're following? Also, don't know how common linters are for C++, but that would be really helpful to add.

afritz1:
I'm following my own style for the subset of C++17 I've chosen. It is not so different from the Visual Studio default. There is some info in the Contributing file I think.

Some of my code in that render() feels pretty old. Any time `TextureUtils::generate()` is called, that allocates a new texture, which is kind of iffy when done every frame, but it's an options menu (shrug).

afritz1:
Done and ready to merge. I've been trying to use `std::optional` more often these days as it's easier to understand than `-1` or a pointer.

Issue #145, opened 2019-04-05
Deadlock in renderer on Linux (and rarely Windows)


afritz1:
When in the game world, the game will randomly freeze on Linux. It appears to be a deadlock with one of the render threads. This behavior is strange because I haven't seen any deadlock issues on Windows with the same code.

Tested on Ubuntu 18.04 64-bit.

afritz1:
Appears to not deadlock if only using one render thread (render threads mode = very low). Anything above that has a risk to freeze.

Carmina16:
It froze for me on Windows once, with the default settings.

afritz1:
Okay, it's probably a programming error then. I think it might've frozen for me too in Windows, but very rarely (i.e., once every other week).

Translator5:
What time passed till it freezes?
For me everything works well, playing it for several minutes (~ 10 minutes), using default settings.

afritz1:
On Ubuntu with 8 threads, usually anywhere from 2 seconds to a minute maybe. Can't really reproduce it on Windows with any thread count.

Translator5:
> On Ubuntu with 8 threads, usually anywhere from 2 seconds to a minute maybe. Can't really reproduce it on Windows with any thread count.

Nope, for me it works fine - as I wrote, 10 Minutes without freezing.

GregTheMadMonk:
For me, game seems to become playable even with max threads settings when I replace 
```
threadData.condVar.wait(lk, [&threadData, &data]()
{
	return data.threadsDone == threadData.totalThreads;
});
```
(line 6357 of SoftwareRenderer.cpp) with just
`threadData.condVar.wait(lk);`
Idk why yet

afritz1:
I'm not sure. Maybe it's an issue with spurious wake-up that's causing a race condition within the `wait()` lambda (although the lambda there is _designed_ to guard against spurious wake-up). There must be a race condition somewhere or else the deadlock wouldn't happen.

I don't think the `threadsDone` variables need to be atomic, but maybe they do.

afritz1:
I ran the GCC thread sanitizer on the engine and, while it found a couple data races (that turned out to be false positives, I'm 99% sure), it did not report anything when the software renderer froze. So I'm still stumped on this issue.

afritz1:
Talked on Discord a bit about this, and it looks like a better idea would be to replace the existing multi-threaded rendering code with a different worker thread design.

jdougharty:
[...]

PR #199, opened 2021-02-08
SoftwareRenderer: Fix data race on flats.threadsDone

Alucowie:
When the last render thread notify that flats are drawn, the main thread
have time to loop on and init flats again before others render threads
have time to do their own checks and continue, so they stay waiting
because flats.threadsDone was init to 0 again.

Remove the wait is safe because the first render threads to finish will
wait on the threadData.go already to false.

### DIFF
--- a/OpenTESArena/src/Rendering/SoftwareRenderer.cpp
+++ b/OpenTESArena/src/Rendering/SoftwareRenderer.cpp
@@ -7856,17 +7856,9 @@ void SoftwareRenderer::renderThreadLoop(RenderThreadData &threadData, int thread
 			{
 				lk.unlock();
 				threadData.condVar.notify_all();
+				return;
 			}
-			else
-			{
-				// Wait for other threads to finish.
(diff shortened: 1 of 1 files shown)


Alucowie:
I think this patch fixes issue #145 .

afritz1:
I tested this on Windows and Raspberry Pi OS and I don't see any issues. I thought it would feel more liberating having a two-year-old bug fixed but it essentially just brings the Linux performance up to where Windows has been, which is still pretty poor in the big picture.

Anyway, thanks a lot for fixing that.

Digital-Monk:
Just wanted to put in another THANK YOU for this.  I enjoy running around Tamriel as a break from work.  My work machine has 12 cores, and being able to use all of them (as opposed to 1 before) makes a huge difference in performance for me.  (As for "still pretty poor", well, I'm happy with anything around 20fps for these kinds of games, so maybe I'm easy to please.)

PR #210, opened 2021-06-24
Changing install instructions to an ordered list

Thunderforge:
Instead of an unordered list (bullet points), the instructions are now an ordered list (1, 2, 3) since they are sequential instructions that need to be followed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,12 +44,12 @@ Check out [CONTRIBUTING.md](CONTRIBUTING.md) for more details on how to assist w
 If you would like music played in-game, see **Music setup** below. The engine uses `ArenaPath` and `MidiConfig` from the options file to find where the game files and MIDI config are.
 
 ### Windows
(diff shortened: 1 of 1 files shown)


afritz1:
Don't know why I didn't think of this. It's weird that all the numbers in your changes are 1 though but maybe that's how Markdown works.

Thunderforge:
> It's weird that all the numbers in your changes are 1 though but maybe that's how Markdown works.

In Markdown, the exact value is actually ignored and it is sequential from whatever the first number is. So if you do:

```
5. Foo
7. Bar
11. Baz
```

then it renders as

5. Foo
7. Bar
11. Baz

A common practice is to make them all ones so that it starts at number one and then the rest don't matter. This is convenient if you change instructions later; if you eliminate step 4, then you don't need to renumber them all.

afritz1:
Okay.

PR #212, opened 2021-06-24
Using shield icons for readme info

Thunderforge:
Moving version, license, Discord, and YouTube info into shields.

Shields

Note that Discord won't work unless [the Widget setting on the server is enabled](https://vimeo.com/364220040).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # OpenTESArena
 
 [![GitHub release](https://img.shields.io/github/release/afritz1/OpenTESArena/all.svg)](https://github.com/afritz1/banner/OpenTESArena/latest)
-[![GitHub](https://img.shields.io/github/license/afritz1/OpenTESArena)](https://github.com/afritz1/OpenTESArena/blob/master/LICENSE.txt)
+[![MIT License](https://img.shields.io/badge/license-MIT-green)](LICENSE.txt) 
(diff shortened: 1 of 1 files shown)


Thunderforge:
If it's too busy, we could probably remove the MIT license since it appears prominently in both the desktop and mobile view. The releases are also visible, but you have to scroll to the bottom in the mobile view.

## Desktop
image

## Mobile top
image

## Mobile bottom
image

afritz1:
Hmm neat. I wonder how well that would work with forks. I think you need to replace full URLs with just the local file in that case, like
```
[...](https://github.com/afritz1/OpenTESArena/blob/master/LICENSE.txt)
```
becomes
```
[...](LICENSE.txt)
```
I do it with the screenshot I think.

Thunderforge:
I just checked and that does work to have a local file reference, but only when clicked from the Readme. If you click on it in this PR, it won't work.

The license image itself was dynamically queried from the repo. However, since the MIT license won't be changing any time soon, I can change it to a static label like so:

```
[![MIT License](https://img.shields.io/badge/license-MIT-green)](LICENSE.txt) 
```

[MIT License](LICENSE.txt)

afritz1:
Okay cool. I'll turn on the Discord widget too.

afritz1:
The `banner` part of the release shield seems to break the link.

Thunderforge:
> The `banner` part of the release shield seems to break the link.

Fixed in #213

Issue #188, opened 2020-10-05
Add support for buffer reference in buffer view constructors


afritz1:
The buffer view classes in [components/utilities/](https://github.com/afritz1/OpenTESArena/tree/fb9db1509f83f18a8c630cc3acee23a693ce07d7/components/utilities) (BufferView, BufferView2D, BufferView3D) are currently a little underutilized and inconvenient, and I would like them to work more naturally with the buffer classes (Buffer, Buffer2D, etc.). A problem I'm facing with that though is related to const-ness.

I have tried and tried all kinds of const/not-const, `std::enable_if_t`, and `std::is_const_v` on the buffer view constructors and functions but I still get compiler errors when I try to do things like this:
```C++
#include "Buffer.h"
#include "BufferView.h"

class MyClass
{
    Buffer<int> buffer;

    const BufferView<int> getBufferView() const
    {
        const BufferView<int> bufferView(this->buffer); // error: Buffer's T is const, BufferView's T is not.
        return bufferView;
    }
}
```
I don't want `T` to have to be `const` but currently a lot of my buffer views are like `const BufferView<const int>`. This feels like it's brushing up against similar issues that `std::vector` has with `const_iterator`.

I want this constructor to exist so I can make a `const BufferView<T>` from a `const Buffer<T>`:
```C++
BufferView(const Buffer<T> &buffer)
```

I wonder if some mixture of `std::underlying_type` or `std::remove_const` or something like that would help. I've given up at this point :/

afritz1:
I think the source of the problem is that `BufferView::data` is `T*`, so if we just do `std::remove_const` on the `Buffer::get()` in `BufferView(const Buffer<T>&)`, then we could assign the pointer. This breaks the rules a little bit but the rules will be upheld by all of the public BufferView functions.

EndlessVoid1:
Have you made any progress on this?

afritz1:
No, have not looked at it, focusing on making the release.

afritz1:
C++20 `std::span` is similar and seems to use `std::remove_cv_t`. Maybe that would be useful here.

makemeunsee:
I tried using `span` directly, and indeed it looks useful; some bits could be simplified.
I havent found an actual usage of any `Buffer` constructed with offsets, so I havent tried but it should be as straightforward as calling `.subspan(viewOffset, viewLength)`.
Here's is a diff showing the usage of `span` replacing `BufferView` in a few select case:
```diff
[...]

PR #217, opened 2021-07-30
#188 - provide read only BufferViews

makemeunsee:
somewhat extensive changes but mostly straightforward:
* created read only versions of `BufferView` and `Buffer2DView`
* used the read only versions everywhere I found and it made sense
* changed refs to move semantics in a handful of cases

of note, [this code](https://github.com/afritz1/OpenTESArena/compare/main...makemeunsee:188-bufferviewreadonly?expand=1#diff-bf598bec18f69d37f1387b9b29935a9a66ac9b43661ecc9e82e745228f7de688R1971) doesnt seem called ever, maybe I miss something?

### DIFF
--- a/OpenTESArena/src/Assets/LGTFile.cpp
+++ b/OpenTESArena/src/Assets/LGTFile.cpp
@@ -26,9 +26,9 @@ bool LGTFile::init(const char *filename)
 	return true;
 }
 
-BufferView<const uint8_t> LGTFile::getLightPalette(int index) const
+BufferViewReadOnly<uint8_t> LGTFile::getLightPalette(int index) const
 {
 	DebugAssert(index < LGTFile::PALETTE_COUNT);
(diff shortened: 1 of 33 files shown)


afritz1:
I need some time to think about this. I'm not immediately satisfied with the `not const ...ReadOnly&` pattern I'm seeing. It just looks weird, but it's that or a redundant `const ...ReadOnly&`. I think I would prefer the latter. Removing `const` from the template parameter seems like an improvement.

I think the changes from copy to move are unnecessary.

The wildBlockIDs are referenced in `MapDefinition::initWild()`.

afritz1:
Saw a video recently where the person stored both non-const and const pointer members and used those with non-const and const functions. Haven't looked into it here but might work better?

afritz1:
Ended up doing this another way that seems to work fine, and I improved all the buffer view classes in the process.

PR #220, opened 2021-11-08
UI draw call registration

afritz1:
Previously, all UI panels that wanted to draw something implemented a virtual `render()` or `renderSecondary()` function. Now all UI rendering is done with UI draw calls that are registered with various logic functions in the Panel's `init()` function. This makes each panel less hardcoded, more configurable, and more data-driven because the draw calls are put into a list and executed by the Game loop. Future work could involve certain panels being replaced with reusable UI collections of some kind, but right now I am pretty tired of refactoring.

This branch took three months because I've been busy with work lately, but I'm glad it is getting merged now.


afritz1:
Works fine on Windows and Raspberry Pi.

Thunderforge:
Seems to work fine on macOS too!

PR #222, opened 2021-12-04
Add exclusive fullscreen support

afritz1:
Replaces my first attempt PR #158 from 2019.

It seems to work in general on Windows 10. Will test on Raspberry Pi 4.

Two bugs found so far that I'm okay with merging because they are minor:
1) If an SDL_Texture (specifically the debug profiler text) tries to update when exclusive fullscreen is on but the game is not focused, the lock texture call fails and an assertion is hit in `TextBox::getTextureID()`. This happens when alt-tabbing.
2) Starting the game in exclusive fullscreen and then switching to windowed will make the window appear like borderless fullscreen because the renderer wasn't initialized with the dimensions from the options

### DIFF
--- a/OpenTESArena/src/Game/Game.cpp
+++ b/OpenTESArena/src/Game/Game.cpp
@@ -74,11 +74,17 @@ Game::Game()
 	}
 
 	// Initialize the renderer and window with the given settings.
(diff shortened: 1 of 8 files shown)


afritz1:
Works fine on Raspberry Pi 4.

PR #227, opened 2022-04-27
Change Bethesda.net Launcher to Steam

TotalCaesar659:
Bethesda [will shut down](https://bethesda.net/en/article/2RXxG1y000NWupPalzLblG/sunsetting-the-bethesda-net-launcher-and-migrating-to-steam) its launcher on 11 May and released The Elder Scrolls: Arena on Steam for free, so it's time to change the corresponding URL.


afritz1:
Okay. Given that Steam usually has a predictable location on the player's hard drive, I can see us wanting to automate the directory look-up a bit more in the future.

PR #228, opened 2022-05-20
Change GOG URL

TotalCaesar659:
In honor of GOG sale, Bethesda made The Elder Scrolls: Arena available for free download without the necessary buying some other of their games.


Thunderforge:
Thanks for the change!

afritz1:
Should this line in the Readme be changed?
> There are two versions of Arena: the floppy disk version (which Bethesda released for free) and the CD version.

TotalCaesar659:
@afritz1 Do you mean for this?

>There are two versions of Arena: the floppy disk version and the CD one.

afritz1:
Yeah. I think we want to say something like
> There are two versions of Arena (both available for free): the floppy disk version and the CD one.

Also replace
> The user must acquire...

with
> The player must acquire...

TotalCaesar659:
Alright, created [the PR](https://github.com/afritz1/OpenTESArena/pull/229).

PR #230, opened 2022-05-21
Update URLs to HTTPS

TotalCaesar659:
This PR updates URLs to HTTPS for security reasons.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ If you would like music played in-game, see **Music setup** below after installi
 ### Windows
 1. Get the most recent build from the [releases](https://github.com/afritz1/OpenTESArena/releases) tab.
 1. Get the Arena data files from one of:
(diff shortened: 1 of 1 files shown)


afritz1:
Looks like it works in Chrome. I'm not a web dev so can't confirm it for the general case. The `Download the Full Game` link on the official website goes straight to the Steam page now, hmm. I wonder if this .zip file on the Bethesda website will go away sometime.

TotalCaesar659:
> Looks like it works in Chrome. I'm not a web dev so can't confirm it for the general case.

You can test both URLs in Chrome after pressing F12 --> "Security", it will show download state.

> The `Download the Full Game` link on the official website goes straight to the Steam page now, hmm. I wonder if this .zip file on the Bethesda website will go away sometime.

I suppose this file is present somewhere (other TES fan sites or else), so it would be great idea to add them as mirrors.

PR #236, opened 2022-11-14
Update setup_windows.md

gazorpo:
hello i tried running this command but it didnt work until i added the .\


afritz1:
I never use PowerShell but I'm not really seeing examples online that require `.\`. Maybe `vcpkg` needs to be added to your Path environment variable.

gazorpo:
ok do you want me to add instructions to add it to Path environment variable then?

PR #237, opened 2022-11-16
Fix slippery ground

gazorpo:
hello i noticed the player character slide around more than the original game when stopping walking so i added some extra friction in that situation, hope it helps


afritz1:
I'll merge this but it's probably going to be replaced later. All the movement code will get rewritten once there's a pass done w/ player attributes (i.e. speed) and collision physics. Right now it's just debug movement.

The original game doesn't have friction for slowing down, you just stop immediately, and there's a bug where it uses the mouse cursor position to influence that stopping.

PR #238, opened 2022-11-17
add character attributes (strength, intelligence, etc)

gazorpo:
im not experienced in c++ so i wasnt sure about when to use const or pointers or references and there is some spaghetti code around the race and gender based attribute generation so if you have any tips there or anywhere else really then im open to it 😊

oh and this doesnt let you increment the stats it just generates and shows them, doesnt even do anything with them in terms of affecting player actions

and btw i just used uesp for info on how to generate stats because im a noob at exe decomp but uesp has sketchy info sometimes so im open to suggestions there too


afritz1:
Overall this is a great contribution, thanks! I'll push some changes to that branch in a while.

afritz1:
Let me know if you're done or you want to add more. I removed the attribute values copied from UESP (mostly to avoid intellectual property issues) since they should come from here instead at some point:
https://github.com/afritz1/OpenTESArena/blob/76c2dfbb96428271b02ad10578dde865f5378ce8/OpenTESArena/src/Assets/ExeData.h#L155

gazorpo:
thank you for fixing my severe lack of references etc 😅 im done, it sounded like on discord you wanted all 0s in attributes for now and i think i need to do some c++ homework before i make more contributions anyway lol

PR #249, opened 2023-05-16
:tada: Added cross-platform CI

AnotherFoxGuy:
Updated GitHub actions to run builds on Windows, Linux and MacOS
This uses Conan for installing the dependencies since it is the easiest way to get the cross-platform builds working
I will open a separate PR for generating artefacts that you can download and run for easy testing


afritz1:
Anything that improves our use of GitHub Actions is a good thing. First time hearing about Conan.

Just glanced through the changes. I think CMAKE_BUILD_TYPE should be ReleaseGeneric because I'm not using the default build types.

Can look more tonight.

Thunderforge:
To be clear: the usage will still keep us within the free tier of GitHub Actions, right? Given that there isn't any budget for this project, I think it would be rejected if it took us to the paid tier.

AnotherFoxGuy:
> The usage will still keep us within the free tier of GitHub Actions, right?

Yes, it absolutely should
From the gh-actions docs: [GitHub Actions usage is free for standard GitHub-hosted runners in public repositories](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#about-billing-for-github-actions) and these builds should be optimized enough to run on the GitHub-hosted runners

afritz1:
I looked this over and don't think it's necessary now, we have builds running for Ubuntu and macOS and I test frequently enough on Windows, WSL2, and Raspberry Pi.

PR #254, opened 2023-10-18
Update image URLs in README.md

theDrake:
Fixed the issue of images not showing up in the README (replaced "master" with "main" in the URLs). Also added the seven "Preview" images and made a minor adjustment to the text.


afritz1:
Thanks. I think the images should be in two groups though. The first four are tech demo examples before things really got started, and the rest are snapshots of the engine's development over time. I can change it.

Issue #201, opened 2021-04-08
Fog rendering and weather


afritz1:
In the original game, fog is an extra layer in screen space that follows the player around and acts like a mask over the scene.
![fog](https://user-images.githubusercontent.com/6729557/113977717-da4b0d00-97f7-11eb-9597-459d3004abc0.png)
I experimented with FOG.TXT a bit. After some trial and error, I got some decent results by treating it as an uncompressed 128x128 image, two bytes per pixel, and dividing each pixel value by 4096 to get an intensity percent.
![screenshot000](https://user-images.githubusercontent.com/6729557/113977820-ff3f8000-97f7-11eb-8e5a-cf79483ea37c.png)
This doesn't seem to be what the original game uses, and it was mentioned on Discord that the game uses a very complex noise algorithm instead.

Allofich:
I don't know about FOG.TXT, but the game does appear to load the FOG.LGT file.

afritz1:
I think FOG.TXT is actually used. Here are some notes from WalterPPK on Discord:
> fog is generated by some very very complicated algorithm (akin to perlin noise?)

> it seems the fog is a flat field around the player

> you somehow project it to 40x25 cell on the screen and do some video memory manipulations with it

> preliminarily, it seems to be a 'cube' of 3000x3000x1000 dimension, and each block is 2x2 pixels on fog.txt map
it will take some time to make the heads and tails out of it: there is self-modifying assembly code and huge unrolled loops

FOG.LGT is a light table that tells what colors to use for darker areas. 13 palettes for 13 light levels.

Carmina16:
As a temporary implementation, you can generate a random 40x25 array of 0..`MAX(FOG.LGT)>>8`, zero out the 9th row, and draw a linear gradient for each 8x8 pixel block on the screen. Maybe dithering should be used to.

afritz1:
I will try that soon. I am figuring out rain and thunderstorms right now, it seems fairly easy except for the lightning bolt generation.

Allofich:
> I think FOG.TXT is actually used.

To test, I modified FOG.TXT in my directory and yes, it definitely is used. Modify the contents and you'll see artifacts appear in the fog in-game.

afritz1:
Need to add TXT file support to the texture manager (ironically). It's not really a text file.

Carmina16:
I described lightning effects in the wiki.

afritz1:
[...]

PR #259, opened 2024-04-22
Data offset corrections

Allofich:
Fixes the offsets for `SpellcastingItemAttackSpellPricesPerCharge`. They were pointing into the previous data at `SpellcastingItemAttackSpellQualities`.

Here's the data in A.EXE (after decompression). 

image

Each set of data has 15 entries for the 15 item attack spells. The value for the start of `SpellcastingItemAttackSpellPricesPerCharge` was pointing at 0x3D5E2, which is 4b80:1de2 here, within the `SpellcastingItemAttackSpellQualities` data.


Allofich:
`SpellcastingItemMiscSpellPricesPerCharge` was similarly wrong. Here is the data from A.EXE.

image

Each set of data has 8 entries. The value for the start of `SpellcastingItemMiscSpellPricesPerCharge` was pointing to 4b80:1f1c here, halfway through the data.

Allofich:
`WaterTwilightLightLevels` is probably incorrect, too, for A.EXE at least (I haven't checked ACD.EXE). The offset points into the middle of a two-byte value. It should probably be one byte less, at 0x3F40A.

From A.EXE:
image

The current value of 0x3F40B points to 4b80:3c0b here (subtract 0x3B800 to get the value shown in the 4b80 segment here). The data starts at 3c0a, though, and 3c0a is what the code references, not 3c0b.

image

Allofich:
src/Assets/ExeData.hpp has the following line

std::array<uint16_t, 14> waterTwilightLightLevels;

but it looks like, starting from the corrected offset, it has 16 elements, the last two being 0.

image

Allofich:
The `spellcastingItemMiscSpell` data had the wrong number of entries set in `ExeData.h` (9 instead of 8) for the values except for the spell names.

afritz1:
Thanks Allofich. I'm checking this branch but I think the spell prices per charge need to be two bytes each. Otherwise `spellcastingItemAttackSpellPricesPerCharge` is like `[0x32, 0x0, 0x32, 0x0, 0x64, ...]`.

Allofich:
You're right. I'll fix that as well.

afritz1:
I got it, no worries!

I'm looking at the water light level values but it seems okay at a glance. Will test some more. I'm pretty sure the game only has 14 light levels, so two extra ones might not make sense.

Allofich:
`armorEnchantmentBonusPrices`, `weaponEnchantmentBonusPrices`, `spellcastingItemBasePrices` and `enhancementItemBasePrices` should also be read as 2-byte values.

afritz1:
Done.

Yeah so about the light levels, if the water changes during sunrise/sunset like I'm thinking, then the values in that array should be between 1 and 13. They're probably indices into which .LGT light table to use to convert texture colors to shaded colors. See these in `ArenaRenderUtils.h`:
```C++
	constexpr uint8_t PALETTE_INDEX_LIGHT_LEVEL_LOWEST = 1;
	constexpr uint8_t PALETTE_INDEX_LIGHT_LEVEL_HIGHEST = 13;
```

Allofich:
Well, if counting from the adjusted offset I suggested the final four 00 bytes seem to fit neatly in the array as two "0" values, and also the values are in descending order by 100h, so they fit the pattern. It could be though that the game never accesses the water light level values with an index high enough that it would reach them.

afritz1:
I changed it. Maybe Arena treats them like # of bytes to skip. If there's nothing else then I'll merge this now.

If you're looking for anything else to help with, it would be awesome to have the screen-space fog finally documented in the wiki in a concise way. I haven't touched it since Carmina gave the info in #201 and honestly it still scares me a little (with the apparent self-modifying assembly from what I heard).

Allofich:
`WaterTwilightLightLevels` (and maybe `WindowTwilightColors`) seem to be related to in-game minutes.

Here is the only place in the code (as far as I can find) that references them:

```
MOV        BX,word ptr [GameMinutesPassedToday]
JS         LAB_1000_04b9
SUB        BX,0x100
JMP        LAB_1000_04bf
LAB_1000_04b9
SUB        BX,0x300
NEG        BX
LAB_1000_04bf
MOV        SI,BX
ADD        BX,BX
MOV        BX,word ptr [BX + WaterTwilightLightLevels]
CMP        BX,word ptr [DAT_4b80_3b4b]
JZ         LAB_1000_048a
MOV        word ptr [DAT_4b80_3b4b],BX
PUSH       BX
MOV        BX,SI
ADD        SI,SI
ADD        SI,BX
ADD        SI,WindowTwilightColors
```

You can see that in 
MOV        BX,word ptr [BX + WaterTwilightLightLevels]
the value in in BX, which was `GameMinutesPassedToday` with various adjustments, has the value of WaterTwilightLightLevels added to it.

`GameMinutesPassedToday` is the value at `4b80:cb70` in the disassembly of A.EXE. I named it that based on its usage in the function where the MIDI to play is determined, as it's compared against in order to decide whether to play the night MIDI or not.

```
MOV        SI,s_night.xmi_4b80_40e2                         = "night.xmi"
 CMP        word ptr [GameMinutesPassedToday],0x10c          = 0100h
JBE        LAB_311a_056a
CMP        word ptr [GameMinutesPassedToday],0x2f4          = 0100h
JNC        LAB_311a_056a
MOV        SI,s_sunnyday.xmi_4b80_40ec                      = "sunnyday.xmi"
```

It would presumably be 0 each midnight, then start counting up from there until resetting to 0 at the next midnight.

Looking at the values of `WaterTwilightLightLevels`, though, the first one is A00h, or 2560, which is more than the number of minutes in a day, and doesn't divide neatly into hours or anything (it's 42.66666666666667 hours), which shakes my confidence a little bit. But, tentatively, I would guess `WaterTwilightLightLevels` is related to in-game minutes.

Allofich:
> If you're looking for anything else to help with, it would be awesome to have the screen-space fog finally documented in the wiki in a concise way.

I may take a look at it, but if it's complicated it may be beyond me.

As you can see in the code snippet above, the values from `WaterTwilightLightLevels` are used to compute a value stored in `DAT_4b80_3b4b`. `DAT_4b80_3b4b` is used in a few different places. It is sometimes compared against 0xFFFF, and it is stored in BX right before a call to a function that references "waterani.rci", so it's probably used with that, which maybe you already know.

Allofich:
> If there's nothing else then I'll merge this now.

I don't intend to add any more commits.

Allofich:
Sorry, the value I called `GameMinutesPassedToday` isn't game minutes, it is 1/1024ths of the day that have passed. (Looking back at the MIDI-related code I noticed that the values didn't match the game if they were interpreted as minutes). So the `WaterTwilightLightLevels` would be related to that.

PR #272, opened 2025-05-02
Bump Jolt to 5.3.0 since it brings PowerPC support

Link4Electronics:


afritz1:
Played around a bit, seems equally as stable as 5.1.0. You shouldn't have bothered Jorrit with adding support for your Mac though lol, you're the only person using it.

afritz1:
Wondering if this Illegal Instruction error in `JPH::sAddVertex()` on my RPi 4 is from 5.3.0, it worked fine in 5.1.0. Seems like -Ofast could be exploiting undefined behavior in Jolt maybe. -O3 works.

Issue #271, opened 2025-04-26
Assign attribute points in character creation


afritz1:
The `ChooseAttributesPanel` screen only shows numbers before save/reroll and doesn't do anything yet.

<img width="1280" alt="Image" src="https://github.com/user-attachments/assets/e8a36ff2-cc32-4cf3-9302-fe63dbfab842" />

It needs the up/down arrow `ArenaTextureName::UpDown` and eight buttons over the attributes for selecting them as the active one.

It also needs the UI for points to allocate: `ArenaTextureName::BonusPointsText`.

When a point is added or removed in the button callback, the `attributeTextBox` needs to combine the base attribute points with the changed points and call `setText()`. The changed points could be an `int[8]` in `CharacterCreationState`.

When confirming character creation and exiting to the game, `ChooseAttributesUiController::onSavedDoneButtonSelected()` needs to include the changed points for `player.init()`.

ChooseAttributesPanel
https://github.com/afritz1/OpenTESArena/blob/7c32b7a02869098bb4d6b2a8a8bb492b99c1add7/OpenTESArena/src/Interface/ChooseAttributesPanel.h#L12

CharacterCreationState
https://github.com/afritz1/OpenTESArena/blob/7c32b7a02869098bb4d6b2a8a8bb492b99c1add7/OpenTESArena/src/Player/CharacterCreationState.h#L11

Initialize player
https://github.com/afritz1/OpenTESArena/blob/7c32b7a02869098bb4d6b2a8a8bb492b99c1add7/OpenTESArena/src/Interface/CharacterCreationUiController.cpp#L542

PR #273, opened 2025-05-06
Assign attribute points in character creation #271

libertus21:
image

Added the UI of the up/down arrows to the character creation panel. Click on an attribute to select it and display the arrows.

Added the UI of the bonus points button.

Up arrow increases attribute value by spending bonus points, Down arrow decreases attribute value and recovers spent bonus points.

The reroll stat button generates again bonus points in a random way.

Range of possible starting bonus points is 4-24


afritz1:
Thank you. I'll review and add some changes then merge tomorrow most likely.

PR #275, opened 2025-05-11
Setup automation

libertus21:
I made this .bat to compile the OpenTESArena code in an automated way without typing the commands by hand, it is based on this guide:

https://github.com/afritz1/OpenTESArena/blob/main/docs/setup_windows.md

I only made it for windows. If approved, it would be nice to make the equivalent for Linux and Mac.

The bat procedure is:

1) Download the necessary tools, vs2019,Cmaker,vcpkg and the necessary packages.

2) Create the Build folder

3) Ask the user what type of build he wants: Debug, ReleaseGenericNoLTO, ReleaseGeneric, ReleaseNative

4) Create the .sln solution with cmake and with an option to also disable certain CPU functions in case of illegal instruction errors.

5) build the project from the command line

6) Copy and paste the data and options folder to the otesa.exe folder.

7) alter options-default.txt to configure the ArenaPaths with the ARENA path inserted by the user and the MidiConfig.

8) Time to program!


afritz1:
Thanks. I need to think about it, there are so many assumptions being made and I don't plan on maintaining this myself.

afritz1:
CMake is now properly copying data and options to the executable folder so you can take that out. Also would you kindly translate the Spanish parts to English? ty

libertus21:
Ah sorry, I overlooked those variables in Spanish. Done, it's already modified

PR #276, opened 2025-05-11
Character stats calculation

libertus21:
Made 3 functions that allow to calculate maximum fatigue, life and magic points.

- calculateMaxStamina: maximum stamina based on Strength and Endurance.
- calculateMaxHealthPoints: base health points using class-specific hit die rolls
- calculateMaxSpellPoints: maximum spell points using class-specific intelligence multipliers


afritz1:
Thanks, tried to keep the important values in CharacterClassDefinition.

PR #277, opened 2025-05-19
Bonus by attributes in choseeAttributePanel 

libertus21:
In choseeAttributePanel now see the and it calculates the bonuses derived from the main attributes which are:  Damage, magic de, to hit, to defend, healh, heal mod, charisma 

image


afritz1:
Thanks, I'll probably merge main into this then add more afterwards.

PR #280, opened 2025-05-23
Basic ai enemy only approaches the player

libertus21:
A very basic AI to enemies. They just follow the player's current location and that's it.

https://github.com/user-attachments/assets/b7293eed-bb9b-4d09-9252-055a5e168c0c


Thunderforge (commented):
I think it would be good to make the enemy AI more modular. Different enemies will need to behave differently (some just attacking, some casting spells, etc) and with this code I think it might be a bit difficult to swap out behaviors.

PR #281, opened 2025-05-24
Updated spell points in chooseAtribute

libertus21:
I noticed that the spell points were not updated UI when the intelligence attribute was raised. Now it does.

The same was true for fatigue points.


afritz1:
Thanks

Issue #279, opened 2025-05-23
Mismatched texture on bottom of wall in start dungeon


afritz1:
In Arena it appears to use *BOXCAP 8 in START.INF which is WALLF.SET for the texture on the bottom of walls.

![Image](https://github.com/user-attachments/assets/c6ca8d66-60f3-4761-a0bd-766bc81223ad)

Not sure how to replicate this as my loading code only knows about one texture ID for all faces of the wall (ignore the different water shading, that is intentional in this engine). This problem also extends to the roof of buildings, it's hard to know what it should be because Arena locks the camera below the top of the Y=1 block.

<img width="1280" alt="Image" src="https://github.com/user-attachments/assets/9d934d8a-9e29-492f-8483-6cd1199ea24d" />

https://github.com/afritz1/OpenTESArena/blob/b320aa80e90ce2bb36c6af09d36a717c173db993/OpenTESArena/src/World/MapGeneration.cpp#L487

```C++
	void writeVoxelInfoForMAP1(ArenaTypes::VoxelID map1Voxel, uint8_t mostSigNibble, MapType mapType, const INFFile &inf,
		const ExeData &exeData, ArenaTypes::VoxelType *outVoxelType, ArenaMeshUtils::ShapeInitCache *outShapeInitCache,
		TextureAsset *outTextureAsset0, TextureAsset *outTextureAsset1, TextureAsset *outTextureAsset2, bool *outIsCollider,
		VoxelFacing2D *outFacing)
	{
		DebugAssert(map1Voxel != 0);
		DebugAssert(mostSigNibble != 0x8);

		if ((map1Voxel & 0x8000) == 0)
		{
			const uint8_t mostSigByte = ArenaLevelUtils::getVoxelMostSigByte(map1Voxel);
			const uint8_t leastSigByte = ArenaLevelUtils::getVoxelLeastSigByte(map1Voxel);
			const bool voxelIsSolid = mostSigByte == leastSigByte;

			if (voxelIsSolid)
			{
				// Regular solid wall.
				*outVoxelType = ArenaTypes::VoxelType::Wall;
				outShapeInitCache->initDefaultBoxValues();
				ArenaMeshUtils::WriteWallRendererGeometryBuffers(outShapeInitCache->verticesView, outShapeInitCache->normalsView, outShapeInitCache->texCoordsView);
[...]

PR #282, opened 2025-05-25
issue-279

libertus21:
issue: #279

Here I have some doubts, is this looking under the wall and it has a different texture exclusive to the starting dungeon, or in other dungeons you can also look under the wall and it has a different texture or it has the same texture in ARENA?

Because if it is exclusive to the starting dungeon and in other dungeons you can't see the texture below the wall then the player wouldn't notice that it is boxcap8 or the same texture as the side wall.

In the code I assign BOXCAP 8 if it is available, otherwise I use the same side wall texture. 

image


afritz1:
We should assume the player can get underneath any wall with a spell like Pitfalls. 

This feels like a texture lookup probably defined in the level somewhere. Maybe in the .MIF or the .INF file, it's not understood yet.

libertus21:
Ok so using WinArena I see that in the walls are defined with sections 

*BOXCAP (base) and *BOXSIDE (side).

walle.set	#3
*BOXCAP 8
*BOXSIDE 5
*DOOR 2

[START.txt](https://github.com/user-attachments/files/20445702/START.txt)

PR #289, opened 2025-07-22
Rasterizer optimization

afritz1:
Converted all the pixel shader code into several `if constexpr` branches in the rasterizer.

Finally was able to get MSVC to produce vector instructions. The sad part is vectorizing did not change performance at 4k at all, which led me to believe it's very memory-bound. Tried changing doubles to floats but it was only a small improvement and caused several rendering bugs to appear, so discarded that.

Fixed puddle rendering by splitting into two passes so it doesn't sync on every single entity and severely slow down rendering.

For now this feels like the upper limit of rendering performance until a graphics API is integrated.


afritz1:
Tried several more things like organizing helper functions for vector instructions and changing light bin sizes to rasterizer bin sizes and changing loop unroll from 4 to 8 but they basically all made FPS worse. Disappointed.

PR #296, opened 2025-10-04
Remove key from inventory on use

### LINKED ISSUE #294: Key doesn't disappear when opening a locked door
You might be aware of this already, but when opening the locked door at the start of the game in OpenTESArena, the ruby key picture doesn't disappear from the screen. In the original game it does, and it also disappears from the player's possession.

Allofich:
Fixes #294.

Any reason you want to keep the key in the inventory for now?


afritz1:
This is good, thank you. Thinking about shifting other keys down in the key inventory.

PR #297, opened 2025-10-04
Determine enemy gold amounts like original game

Allofich:
Addresses the todo in `GameWorldUiController.cpp` by determining gold chance for enemies according to the original game. It also corrects the amount of gold generated.

Gold is still generated at the time an enemy corpse is clicked on, the same as before this PR. In the original game, gold is calculated at the time an enemy is instantiated. This means if you save a game with an enemy near you and then load that game, the gold or other treasure they have won't change. I put in a todo note about this.

I confirmed the behavior here matches the original game (version 1.06), but it looks like a mistake to me. Every monster has a set chance for gold and other items to appear (this PR only handles gold). This is rolled against twice, like:

```
("random_range" is exclusive of the max value)
1st roll: If random_range(1,101) <= goldChance, then do the 2nd roll.
2nd roll: If random_range(0,101) >= goldChance, then add gold.
```

Since not only do two rolls happen, but the direction of the second check is reversed, there is a low chance for monsters to have gold, especially for higher-level monsters with high goldChance values. For example, the goblin has a goldChance of 25, so basically a 0.25 x 0.75 chance, or 0.1875% chance for gold. High-level monsters have values like 90, so 0.90 x 0.10, or only a 0.09% chance for gold.

The "LootChances" that are read in from the executable to get the goldChance are 4-byte values, with the least-significant bytes being the gold chances, and the other three being for items. The items don't have to pass a second roll, they just use the `random_range(1,101) <= chance` calculation. My guess is gold probably was supposed to be the same.

At some point I would like to check things like this against version 1.07, the CD version, just in case it has undocumented bug fixes in it.

### DIFF
--- a/OpenTESArena/src/Assets/ExeData.cpp
+++ b/OpenTESArena/src/Assets/ExeData.cpp
@@ -472,7 +472,7 @@ bool ExeDataEntities::init(Span<const std::byte> exeBytes, const KeyValueFile &k
 	initInt8Array(this->creatureBlood, exeBytes, creatureBloodOffset);
 	initInt8Array(this->creatureDiseaseChances, exeBytes, creatureDiseaseChancesOffset);
 	init2DInt8Array(this->creatureAttributes, exeBytes, creatureAttributesOffset);
-	initInt16Array(this->creatureLootChances, exeBytes, creatureLootChancesOffset);
+	initInt32Array(this->creatureLootChances, exeBytes, creatureLootChancesOffset);
 	initStringArrayNullTerminated(this->creatureAnimationFilenames, exeBytes, creatureAnimFilenamesOffset);
 	this->finalBossName = GetExeStringNullTerminated(exeBytes, finalBossNameOffset);
 	init2DInt8Array(this->raceAttributes, exeBytes, raceAttributesOffset);
--- a/OpenTESArena/src/Assets/ExeData.h
+++ b/OpenTESArena/src/Assets/ExeData.h
@@ -139,7 +139,7 @@ struct ExeDataEntities
 	uint8_t creatureBlood[24]; // Indices into effects animation list.
 	int8_t creatureDiseaseChances[24]; // Negative values have special meaning.
 	uint8_t creatureAttributes[24][8]; // 255 == 100.
-	uint16_t creatureLootChances[24];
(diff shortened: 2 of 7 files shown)


Allofich:
Added code to use the rest of the lootChances value to get the chances for non-gold inventory items. The actual items being added are still random (and no enchanted items), but the chances and categories should match the original game.

Allofich:
> Just double-checking -- the `creatureIndex + 1` you're doing in EntityDefinition.cpp is going to make rats get the goblin loot slot, and would cause an out of bounds error with creatureIndex 24 if we didn't have the final boss check. Is that what you wanted?
> 
> I could see there being a Fix Gold Chances game setting so it uses the chances correctly.

Yes, I traced the original game's execution and it accesses the data using the creature type/race, which is 1-based, as the index. That's equivalent to `creatureIndex + 1` here. This makes the 0 slot unused. The values that are gotten look appropriate to the creatures they are gotten for if you line them up side-by-side, so I don't think the rat is mistakenly getting the goblin's data, etc. if that's what you meant.

The check with creatureIndex 24 is to stop the out-of-bounds error, yes. I don't know what the original game does in that case, as I don't see a guard in the disassembly there against accessing the loot chances data in that case.

> I could see there being a Fix Gold Chances game setting so it uses the chances correctly.

Sure. There are going to be lots of things like this, though, so you could end up with a lot of such settings. Maybe eventually you could have a selection of settings that imitates the original game, and one that turns on all fixes. There could be more than one way to approach fixing an issue, though, (not just fixed/not fixed) so that could complicate matters. You could also have a set of settings for the 1.06/Floppy version and the 1.07/CD version (and maybe even for earlier versions if there are any interesting differences). Aside from the added intro and voiceovers, there are a few other differences between 1.06 and 1.07 on UESP noted here: https://en.uesp.net/wiki/Arena:Patches, and there could be more.

Allofich:
Looks like the build failed.

Allofich:
Is `charClassDef.categoryID` in the order thief, mage, warrior? That's the order of the gold chances.

afritz1:
Dang I thought it was mage, thief, warrior like classIndex / 6. I'll fix. Trying to keep CharacterClassDefinition from being glued to the original game.

PR #304, opened 2025-10-18
Make stamina loss like the original game

Allofich:
Includes recreation of the original game's rest healing function, adjustment to the game time scale, and corrections to value scaling in the thieving and bashing functions.

**Stamina Loss**
This recreates the rate of stamina loss from the original game for both being in and out of water. This includes a check against being Argonian, as they lose stamina at a lower rate than other races do in the water.

This includes an adjustment to the game time scale, because stamina drops happen in the original game every in-game minute, and so the rate at which time flows needs to match with the original game. The rate before this PR was 20x real time. Do you know what this is based off of? Testing in DOSBox, I get about 13x real time, even when changing the amount of cycles. For now at least, I changed it to 13x real time.

The stamina loss isn't 1-to-1 with the original game because the original game lowers stamina every game minute, whereas this is happening every game tick (as it was before this PR), but the rate should be the same.

For the stamina loss rate check against the player being Argonian, I added it to the race definition like you did in the previous PR with the climbing and swimming checks.

**Rest Healing Function**
This also recreates the function used by the original game to heal the player when resting. This is used for an hour of rest, and currently only happens when stamina runs out, since the Camp command isn't implemented yet. It references a byte array in the .exe that represents the heal modifiers per tavern room. In the stamina-ran-out case, it always accesses the array at index 1 and gets a 0 modifier, so there is no effect now. I created a new "Services" section in the exe Data strings for the array as I wasn't sure where to put it in the existing categories. I imagine other numerical values like tavern drink prices or values referenced by other services (haven't looked at those) could go there. Feel free to change it if you don't like it there.

The original game makes a check for the player being the Barbarian class in this function, but it has no effect. I think maybe an AND here was supposed to be an ADD, as that would have an effect. I put in a comment about it.

In every place that the original game calls this function, it puts 1 into AX before doing so. This is represented by the `restFactor` parameter here. Since it's always 1 it will never affect anything, so we could omit it, but it's here for completeness.

The function is different from the other ones I've added to `ArenaPlayerUtils` so far in that it doesn't just compute a value and return it, it modifies multiple player values (health, stamina and spell points). It also should restore attributes, but I've left that as a todo because attributes aren't yet represented with base/current values in OpenTESArena. I'm not sure whether you want functions in `ArenaPlayerUtils` to be able to modify things, but when attribute-restoring is implemented, it will need to be able to reference the changing heal modifier as the player's endurance is restored by each hour of rest.

Also, the has checks against player classes. Right now I just handle those directly, but maybe you'll want to reorganize that.

**Corrections**
When I did the thieving and bashing functions, I didn't notice that the player attributes in TESOpenArena are being stored as 0 to 100 values, not 0 to 256 values like the original game, and I did some conversions wrong because of that. I fixed that here. I also took "attribute" out of the name of the conversion functions because values derived from attributes, like `healMod`, can also need the conversions, not just the 8 attributes themselves.


afritz1:
> Testing in DOSBox, I get about 13x real time

I probably just eye-balled real seconds per game minute in DOSBox. I trust your value better!

> The stamina loss isn't 1-to-1 with the original game because the original game lowers stamina every game minute

i'm fine with this for now.

ExeDataServices section is good, that'll probably be expanded.

The player rest healing could probably be a Player method. I'll do that now.

I moved the 256/100 functions to ArenaStatUtils.

afritz1:
I'm not positive about the actual stamina loss, currently we are summing the awake and swimming changes together which is maybe double what it should be?

Allofich:
Here's what I can tell about the timing of Arena:

The game uses Int 8 to increment a variable. According to sources online this timer tick happens 18.2 times per second, or 18.2065 times per second (both numbers appear in search results, I assume the 18.2 is just rounding).

Every update loop, the game uses the variable to check whether the number of timer ticks since the last time it updated the game clock (which is in in-game minutes) has exceeded 90. If it has, it increments the game clock by one.

1 / 18.2065 = 0.0549254387169418 seconds per tick. Times 90 that is at least 4.943289484524758 seconds per in-game minute. That's close to what I measured in DOSBox, which was that in one minute of real time the game clock advanced 13 minutes. 60 / 5 is 12. I measured in DOSBox multiple times and the game always had advanced by 13 game minutes, though, not 12. I don't know if that's because there is delay somewhere or my measuring was slightly inaccurate or what.

PR #308, opened 2025-11-04
Progress on recreating Arena item generation

Allofich:
This covers lots of things but they are all related to item generation.

First, there are some corrections to existing things:
1. For loot generation, I had made the code get the number of items in a loot object before generating them. What the original game actually does is roll for each of four slots (gold, magic item, non-magic weapon/armor, magic weapon/armor), and set a value to 1 or 0 for each of these four slots depending on if the roll succeeded. Then, for each slot that is 1, it generates an item. This PR recreates this behavior.
2. The weapon generation (which wasn't hooked up to anything yet) was picking a random ID, then if that didn't satisfy the quality requirement it would iterate forward through the rest of the weapon list until it found one that did. What the original game actually does is keep picking random IDs from the range of possible IDs until one that meets the quality requirement is met. This PR recreates this behavior.
3. The ItemLibrary used `armorNames` ("Cuirass", etc.) to get the names for plate armor items and had `plateArmorNames` commented out with a comment that you weren't sure "ordinary" plate exists in the game. The original game uses `plateArmorNames` ("Plate Cuirass", etc.) for plate armor without a material bonus. This PR gets the names from `plateArmorNames` now and has commented out `armorNames` until materials are handled.

Next, the new things:
1. Gets non-magic weapons and armors for loot objects like the original game.

2. During this process the original game calls a function to get the item quality, but it doesn't use the result, overriding it with 16. The function is also called during magic item generation, which isn't implemented yet. The function is included in this PR for later use with magic item generation.

3. In the process of getting the non-magic weapon or armor, the original game refers to a set of values that looks like [0, 0, 1, 1, 0], which is accessed using the lootValueIndex. A 0 means to generate with either 25% condition, 75% condition or 100%, like with items on creatures, while a 1 means to generate with either 75%, 100% or 100% (2/3 chance of 100%). Effectively this means that these weapons and armor found in places with index 2 or 3 (NOBLE or DUNGEON) are more likely to have good condition. Since item condition isn't handled yet this isn't hooked up, but the code is there.

4. Added some simple functions to use the itemIDs returned by the Arena functions to get items from the OpenTESArena ItemLibrary via the item display names. Feel free to replace these if you want. While making them I discovered that there is a discrepancy in shield names. The armor lists for leather, chain and plate use "Kite shield" and "Tower shield" with a lowercase "s," and this (specifically, the plate list) is what the original game uses for getting the names of these items when generating a non-magic armor/weapon (BTW, the non-magic armor/weapon generation also means no material bonus). Meanwhile the `armorValues` list, which is used to create names of material plate armor, uses "Kite Shield" and "Tower Shield". ItemLibrary is getting the names for shields from this list, so in the itemID-to-name function, I replace "shield" with "Shield," since I am reading the names from `plateArmors`, to match the names in ItemLibrary. I used these functions to hook up non-magic weapon and armor generation to item generation on creatures and in loot. Only plate armor will appear (as explained in https://github.com/afritz1/OpenTESArena/pull/301). Any leather or chain items that appear are due to the test placeholder for magic weapons and armor, not these functions. The non-magic weapons/armor that appear on creatures are not 1-to-1 with the original game because they aren't affected by the can-equip check.

### DIFF
--- a/OpenTESArena/src/Entities/ArenaEntityUtils.cpp
+++ b/OpenTESArena/src/Entities/ArenaEntityUtils.cpp
@@ -3,6 +3,7 @@
 #include "ArenaEntityUtils.h"
 #include "../Assets/ArenaTypes.h"
 #include "../Assets/ExeData.h"
+#include "../Items/ItemDefinition.h"
 #include "../Math/Random.h"
 #include "../Stats/CharacterClassLibrary.h"
 
@@ -190,7 +191,8 @@ int ArenaEntityUtils::pickNonMagicWeapon(int weaponLevel, int specifiedItemID, c
 	return itemID;
 }
 
-void ArenaEntityUtils::getCreatureNonMagicWeaponOrArmor(int creatureLevel, const ExeData &exeData, Random &random, int *outWeaponOrArmorID, bool *outIsArmor)
+void ArenaEntityUtils::getCreatureNonMagicWeaponOrArmor(int creatureLevel, const ExeData &exeData, Random &random, int *outWeaponOrArmorID,
+	bool *outIsArmor, ArmorMaterialType *outArmorMaterialType)
 {
 	int itemID = -1;
 	bool isArmor = false;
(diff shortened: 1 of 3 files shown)


Allofich:
I'm mostly just doing this as a hobby, as I get satisfaction out of the mental exercise and seeing the game slowly come to life. But I'd say don't feel like you have to quickly respond to my PRs or even keep the project going if you get tired of it. Who knows, in a few years we might be able to just ask an AI to finish the whole thing for us.

PR #310, opened 2025-11-08
Generate magic items like original game

Allofich:
1. Adds the various magic items (consumables, trinkets and accessories, not magic weapons or armor) with their associated enchantments to the ItemLibrary, and generates them for creatures and loot piles like the original game. Items use their identified names currently. Spellcasting items still need to have their number of charges generated, which is random per instance. The number of charges determines the item price, so they don't have their price set yet. Attribute-enhancing and armor class items do have their price set.

2. Removes the message about item tables not being implemented from the loot subpanel. I don't know what "item tables" means here, so maybe removing the message is premature, but the loot subpanel is close to matching the original game now so it feels like a "not implemented" message isn't needed any more. Also removed "test" from the "testItemDef" variable name since it's not just random test items now.

3. Added `SpellDefinitionID` and `AttributeDefinitionID` as int aliases to mimic what you did with `ItemMaterialDefinitionID`.

4. Only add items to an inventory if a valid item was found (itemDef is not -1).

If you want to test, comment out

```
if (!validLootSlots.slots[i])
{
    continue;
}
```

in `EntityChunkManager.cpp` and hardcode `quality` to 7 or higher in `getLootMagicItem` in `ArenaEntityUtils.cpp`. This should let every loot pile always generate a magic item.

### DIFF
--- a/OpenTESArena/CMakeLists.txt
+++ b/OpenTESArena/CMakeLists.txt
@@ -503,6 +503,9 @@ SET(TES_SKY
     "${SRC_ROOT}/Sky/SkyVisibilityManager.cpp"
     "${SRC_ROOT}/Sky/SkyVisibilityManager.h")
 
+SET(TES_SPELLS
+    "${SRC_ROOT}/Spells/SpellDefinition.h")
+
 SET(TES_STATS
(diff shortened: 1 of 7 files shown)


afritz1:
I played a little bit, saw an Elven Bracelet!

PR #322, opened 2026-05-29
Set up for spawning guards

Allofich:
This adds:
1. Function from original game to determine if guards appear when committing violence (killing a citizen or casting a projectile spell). Hooked up to killing citizens.
2. Function from original game to determine if guards appear when doing theft (pickpocketing, bashing, stealing from stores). Not hooked up to anything.
3. Functions from original game to determine type, level and number of guards. The type is one of three progressively higher attribute guards that can appear for each city type. Hooked up.
4. Function from original game to pick spawn points for the guards (probably used by the original game any time an enemy needs to be spawned). This is incomplete, as it uses data from MAP1 and FLOR, as well as an array of tiles currently occupied by entities (citizens or enemies) to check for a valid spawn point. Right now it randomly picks a point like the original game, but doesn't do those validity checks (it only checks that it isn't the same tile as the player, which the original game also does). This is also hooked up, but no actual spawn occurs as I don't know how to do that in OpenTESArena or if it is implemented anywhere yet. For now a debug log message appears.

Functions for the MAP1, FLOR and entity location checks are in but unused, and will probably need to be adjusted. In the original game, it checks a 2-byte value in MAP1 at the index of the spawn point coordinates for anything but 0000, and fails if it is. In-game I saw values like:
0000 (valid)
2f2f
e000
a247
a259
2c2c
2b2b
2929
2121
2222
1f1f
in the MAP1 data. Maybe you recognize what some of that means. Do you know why it seems to be checking that 2-bytes are both 0 in what looks like 1-byte data?

For FLOR, it checks that the 2-byte values at the index of the spawn point is less than or equal to a value that, in-game, was 0xec00. The values I saw in FLOR were all e000, e100 and e200, so those would be valid. Maybe if there had been some water around it would have appeared as a value over 0xec00.

If you know how to pass in MAP1 or FLOR data, or have an idea for keeping an array of positions of nearby entities like the original game (it tracks up to 23 positions and is also used for AI pathfinding to avoid enemies walking into each others' tiles. It corresponds to the `IsTileOccupied` call in https://github.com/afritz1/OpenTESArena/wiki/NPCs#ai), or how to spawn a guard entity, we could apply that to this.

Also, in the original game there is a delay of 15 game update loops before guards appear. I don't know how that translates to OpenTESArena. And also, the spawned guards are spawned facing toward the player. I don't know if you care about reverse-engineering that but it uses `ComputeDirection` as seen in https://github.com/afritz1/OpenTESArena/wiki/NPCs#ai.


Allofich:
Sounds good. Another thing I thought about (and I wonder about for the original game as well), is can guards spawn on the other side of the city walls or behind a palace gate? But I'm fine with considering this PR done.

